PR #44, opened 2026-05-19
feat: built-in @browser command + @-prefix convention for extensions

not-elm:
## Problem

ozmux extensions can register PATH-based command shims (e.g., `memo` → `$OZMUX_BIN_DIR/memo`), but ozmux's own CLI subcommands have no equivalent. A user inside a terminal Activity who wants to open an embedded browser pane has to type the full `ozmux browser <URL>` invocation — the daemon ships every other piece of the feature but exposes no short, discoverable way to invoke it from the shell.

We also wanted a visual convention that distinguishes ozmux-ecosystem commands from regular shell binaries, and to give extensions the same opt-in.

## Solution

Ship `@browser` as a built-in PATH shim materialised by the daemon at startup, plus loosen the SDK regex so extensions may declare `@`-prefixed command names. As a coupled rename, `memo` becomes `@memo` to adopt the new convention.

### How it works

1. **Daemon bootstrap** writes `$TMPDIR/ozmux/<pid>/bin/__builtin/@browser` — a 2-line `#!/bin/sh` script that `exec`s the running `ozmux` binary with `browser` as the first arg (resolved via `std::env::current_exe()`).
2. **PATH ordering**: `TerminalService::extension_path_prefix` now partitions `__builtin` to the head of the PATH prefix explicitly, so built-in shims always win over same-named extensions regardless of ASCII sort order.
3. **Reserved name guard**: bootstrap runs a pre-pass before `ExtensionHandles::load()` and skips built-in materialisation with a loud error log if any extension's `package.json.name` equals `__builtin`. The pre-pass has to run outside `node_handle()` because `ExtensionHandles::load()` swallows per-extension errors.
4. **Self-recursion guard**: `validate_ozmux_exe` rejects baking an `ozmux_exe` path that lives under the runtime bin tree (pyenv-#2696-class bug).
5. **Best-effort policy** for everything in the materialisation chain — every failure is logged and the daemon continues, matching the existing `place_cli_shim` policy.

### Areas touched

- `daemon/bootstrap/` — new `builtin_commands` module + bootstrap wiring (`materialize_builtins`, `check_builtin_name_collision`)
- `daemon/terminal/` — `extension_path_prefix` partitions `__builtin` to head
- `daemon/extension/tests/e2e.rs` — new PTY-driven integration test (`builtin_browser_shim_is_on_path`) verifies the shim resolves on `PATH` via sentinel-wrapped `command -v @browser` inside a real `/bin/sh`
- `sdk/typescript/src/server/shim-writer.ts` — `COMMAND_NAME_RE` loosened from `^[a-z]…$` to `^@?[a-z]…$` (strict superset — every previously-valid name still validates)
- `extensions/memo/bootstrap.ts` — command key renamed to `\"@memo\"`
- `CHANGELOG.md` — user-facing notes and the nushell known limitation

### Design + plan

Architectural rationale lives in `docs/superpowers/specs/2026-05-19-builtin-command-aliases-design.md` (gitignored), reviewed twice via parallel research and updated for must-fix findings before implementation.

### Breaking Changes

- **\`memo\` extension command renamed to \`@memo\`.** Pre-1.0; users who had \`memo\` in shell scripts or muscle memory need to update. CHANGELOG documents the change.

### Known limitations

- **Nushell** uses \`@\` as an attribute syntax for \`def\` declarations and is likely to mis-parse a bare \`@browser foo\` at the prompt. Workaround: \`^@browser foo\` (caret-prefixed external command). Initial release supports bash, zsh, fish, dash, and POSIX sh; nushell empirical confirmation is tracked as a follow-up.
- **brew distribution**: \`std::env::current_exe()\` does not resolve symlinks, so a future brew tap would bake a Cellar-specific path into shims that goes stale after \`brew upgrade\`. Mitigation deferred until brew distribution is on the roadmap.

### Test plan

- [x] \`cargo test\` — 176 Rust tests pass (1 pre-existing unrelated failure on \`foreground_start_and_sigterm_shutdown\`, verified on the base commit)
- [x] \`pnpm test\` — 506 TypeScript tests pass (440 frontend + 66 SDK)
- [x] \`pnpm check-types\` — all packages clean
- [x] \`make fix-lint\` — clean
- [x] Manual daemon smoke test — \`__builtin/@browser\` lands on disk with mode 0500 inside a 0700 dir, first line is \`#!/bin/sh\`
- [ ] Manual UI verification via \`make dev-e2e\` (deferred — requires interactive Playwright MCP session): inside a terminal Activity, verify \`echo \$PATH\` shows \`__builtin\` first, \`command -v @browser\` returns the shim path, \`@browser https://example.com\` opens a browser pane, \`@memo\` still works, and \`@<TAB>\` completion surfaces both built-in and extension commands. Repeat in bash, zsh, and fish.

### DIFF
--- a/daemon/bootstrap/src/lib.rs
+++ b/daemon/bootstrap/src/lib.rs
@@ -3,8 +3,6 @@
 //! (currently the `ozmux` CLI's `daemon start --foreground` command) drive
 //! the tokio runtime themselves and call `run().await`.
 
-/// PID file management for the daemon process: write/read/remove plus
-/// `is_process_alive` and a `PidFileGuard` RAII helper.
 use anyhow::{Context, bail};
 use ozmux_browser::BrowserUnavailableReason;
 use ozmux_browser::cef_registry::BrowserCefRegistry;
@@ -15,10 +13,14 @@ use ozmux_extension::registry::ExtensionRegistry;
 use ozmux_extension::runtime::RuntimeRoot;
 use ozmux_http_server::AppState;
 use ozmux_http_server::activity_titles::ActivityTitles;
+use ozmux_multiplexer::MultiplexerService;
 use ozmux_terminal::TerminalService;
 use std::sync::Arc;
 use std::sync::atomic::Ordering;
 use tokio::signal::unix::{SignalKind, signal};
+
+/// PID file management for the daemon process: write/read/remove plus
+/// `is_process_alive` and a `PidFileGuard` RAII helper.
 pub mod pidfile;
 
 mod builtin_commands;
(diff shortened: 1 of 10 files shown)


daemon/bootstrap/src/lib.rs
```diff
@@ -286,3 +291,140 @@ fn longest_extension_name() -> std::io::Result<String> {
     }
     Ok(longest)
 }
+
+/// Materialises the built-in `@<name>` shims into
+/// `runtime_root/bin/__builtin/`. Best-effort: every failure is
+/// logged and the daemon proceeds without the affected shims. This
+/// matches the policy of `place_cli_shim()` above so the CLI shim
+/// and the built-in shims have consistent behaviour on error.
+async fn materialize_builtins(runtime: &RuntimeRoot) {
```
not-elm: anyhow::Result型に変更して。

not-elm:
PR review feedback (https://github.com/not-elm/ozmux/pull/44#discussion_r3263738046) addressed in `fe57666`:

- `materialize_builtins` signature changed to `async fn ... -> anyhow::Result<()>`
- Body rewritten as a thin `?`-propagator using `anyhow::Context` messages identifying which step failed (`current_exe` / `validate` / `collision` / `materialise`)
- Caller in `run()` now logs at the boundary via `if let Err(e) = ... { tracing::warn!(...) }`, mirroring the existing `place_cli_shim` shape immediately above

Same end-user behaviour preserved (best-effort — built-in shims may be skipped, daemon still starts). The previous per-step `warn!` vs `error!` distinction collapses into a single `warn!` at the boundary; the anyhow chain still surfaces the cause.

PR #64, opened 2026-05-26
Add runtime font override + bundle JetBrains Mono Nerd Font Mono

not-elm:
## Problem

The Bevy GUI bundled JetBrainsMono in the renderer crate, while the UI overlay (status bar, tab bar, copy-mode indicator) fell back to Bevy's default FiraSans/FiraMono — so the terminal grid and chrome rendered in two different typefaces. We also want a way to override individual faces from `~/.config/ozmux/config.toml` for users who prefer a different font, plus a top-level `licenses/` directory that bundles the SIL OFL §1.20 attribution.

### Bundled defaults
- Replaces the four `crates/bevy_terminal_renderer/src/glyph/JetBrainsMono-*.ttf` files with `assets/fonts/jetbrainsmono/JetBrainsMonoNerdFontMono-{Regular,Bold,Italic,BoldItalic}.ttf` — same family, Nerd Font–patched so icon codepoints (U+E000–U+F8FF) render through the grid. Bundled assets total ~10 MB.
- Adds `licenses/jetbrainsmono/{OFL.txt,README.txt}` with the upstream copyright statement (`Copyright 2020 The JetBrains Mono Project Authors`) plus Nerd Fonts patcher credit, per OFL §1.20.
- New `pub mod bundled` in `bevy_terminal_renderer` exposes the four TTF byte slices as `pub const`s. Both `TerminalFonts::default()` and the Bevy app's `FontBridgePlugin` reference these constants instead of duplicating `include_bytes!`, so the linker has a single source of truth and the binary doesn't carry the ~10 MB twice.

### Runtime override path
- `daemon/configs::font::FontConfig` gains four optional `Option<PathBuf>` fields (`normal_path`, `bold_path`, `italic_path`, `bold_italic_path`), all `#[serde(skip_serializing)]` so they never leak into `GET /configs/font` responses. Per-face paths are independent — no inheritance from `normal_path` to bold/italic (raw-path loading has no synthesizer the way OS font lookup does).
- `daemon/configs::path` gains a `pub fn expand_user_path` (handles `~`, `~/`, and `~\` on Windows; refuses non-UTF-8 paths rather than silently mangling).
- New `src/font.rs::FontBridgePlugin` runs at Startup `.before(TerminalFontInitSet::InitCellMetrics)` — a new public `SystemSet` exported from the renderer crate. On the no-override path it short-circuits (only registers the UI `Handle<Font>`); on the override path it loads each face independently and substitutes bundled bytes if a face's path is missing, fails expansion, or fails to parse.
- The bridge inserts `TerminalUiFont(Handle<Font>)` directly via `Assets<Font>::add` (not `AssetServer::load`, which is rooted at `assets/` and can't reach `~/.fonts/...`). On `bevy_text::Font::try_from_bytes` rejection it falls back to `Handle::<Font>::default()` — Bevy's bundled FiraMono — with an `error!` log noting the visible terminal/UI typeface mismatch.

### UI plumbing
- `build_status_bar` / `build_window_are` / `build_pane_tab_bar` / `build_tab` take a `ui_font: &Handle<Font>` parameter and set it on every `TextFont`. Caller systems (`rebuild_structure_on_change` in `src/ui.rs`, `build_cell_recursive`/`build_pane` in `src/ui/layout.rs`) read `Option<Res<TerminalUiFont>>` and pass the cloned handle in.
- `attach_indicator_to_activity_host` (a Bevy system) takes `Option<Res<TerminalUiFont>>` directly; `Option` because the existing test harnesses don't add `FontBridgePlugin`.

### Tests
- `daemon/configs::path::tests` — 5 new unit tests for `expand_user_path` (absolute pass-through, `~/`, bare `~`, no-home, `~user/...`), plus 2 unix-gated non-UTF-8 tests and 1 windows-gated `~\` test.
- `daemon/configs::font::tests` — 4 new unit tests for path-field parsing, default-state, no-inheritance, per-face independence.
- `crates/bevy_terminal_renderer::glyph::font::tests` — pre-insert preservation test verifying `TerminalFontPlugin::build` does NOT overwrite an already-present `TerminalFonts`. Empirical 12 px metric ranges measured against JBM Mono (`advance_phys ~7.2`, `line_height_phys ~15.84`, `ascent_phys ~12.24`, `descent_phys ~3.6`).
- `src/font.rs::tests` — 6 integration tests for the bridge: bundled default, strong-handle insertion, override path, missing-file fallback, no-inheritance, corrupt-bold-doesn't-drop-normal. The test module uses an `EnvVarGuard` RAII drop guard for panic-safe `OZMUX_CONFIG` cleanup.

Total: 127 tests pass across the workspace. `cargo build` is clean.

### Files affected
- New: `assets/fonts/jetbrainsmono/` (4 TTFs, ~10 MB), `licenses/jetbrainsmono/{OFL.txt,README.txt}`, `crates/bevy_terminal_renderer/src/bundled.rs`, `src/font.rs`.
- Removed: `crates/bevy_terminal_renderer/src/glyph/JetBrainsMono-*.ttf` (4 files, replaced by the new Nerd-Font-patched variant under `assets/`).
- Modified: `daemon/configs/src/{font.rs,lib.rs,path.rs}`, `crates/bevy_terminal_renderer/src/{lib.rs,glyph/font.rs,glyph/atlas.rs}`, `src/{main.rs,ui.rs,ui/status_bar.rs,ui/tab_bar.rs,ui/copy_mode_indicator.rs,ui/layout.rs}`, `Cargo.toml` (added `ab_glyph` dep, pinned `bevy/default_font` feature).

### Breaking Changes

None for external users. The web frontend's font-config behavior is unchanged (the new `*_path` fields are `skip_serializing` so they don't appear in `GET /configs/font` JSON; the family-name fields the frontend reads are untouched).

### Known caveats

- A pre-existing post-teardown SIGSEGV in `cargo test -p ozmux-gui` (reproduces at commits before this PR began) appears after `test result: ok.` is printed — all assertions still pass. Out of scope here; tracked separately.


crates/bevy_terminal_renderer/src/glyph/font.rs
```diff
@@ -3,18 +3,67 @@ use bevy::prelude::*;
 use bevy::window::PrimaryWindow;
 use ttf_parser::Face as TtfFace;
 
+/// Error returned by `TerminalFonts::from_bytes` when one of the supplied
+/// per-face byte buffers fails `ab_glyph::FontArc::try_from_vec`.
+#[derive(Debug)]
+pub enum FontLoadError {
```
not-elm, line 9: Please use thiserror crate

not-elm: Done in 050e46e — FontLoadError now uses #[derive(thiserror::Error, Debug)] with #[error(...)] and #[source]. Same Display output and Error::source() behavior; tests unchanged.

PR #80, opened 2026-06-02
refactor: dispatch NewSession/FocusSession via EntityEvents

not-elm:
## Problem

`NewSession`, `FocusSession`, and `FocusSessionNumber` were the last shortcut actions still handled **inline** inside the keyboard dispatcher (`src/input.rs`), unlike every other action which dispatches through a Bevy `EntityEvent` + observer. The inline path mutated the `AttachedSession` marker and spawned session entities directly via `Commands`, and needed a per-frame `marker_dirty_this_frame` guard to protect the "exactly one entity carries `AttachedSession`" invariant against Bevy's deferred-command-flush timing — silently dropping the second marker-mutating action in a frame.

## Solution

Route the three actions through `EntityEvent`s + observers, consistent with the existing pane/activity and clipboard actions:

- **New `src/action/session.rs`** owns `NewSessionActionEvent` / `FocusSessionActionEvent` (plus a local `FocusSessionTarget` selector that folds `FocusSession{offset}` and `FocusSessionNumber{index}` into one event) and their observers. The session-spawn and marker-move logic (`spawn_attached_session`, `dispatch_new_session`, `dispatch_focus_session`) moves here from `input.rs`. Wired via `OzmuxSessionActionPlugin` → `OzmuxActionPlugin`.
- **`src/input.rs`**: `execute_action` now just triggers those events (mirroring how `Copy`/`Paste` are dispatched); `dispatch_focused_key` is slimmed — it drops the `session_name_counter` / `sessions` params and the `marker_dirty_this_frame` guard (and its NOTE).
- **`src/bootstrap.rs`** repoints to the moved `spawn_attached_session`.
- The observers re-query the **live** `AttachedSession` marker instead of trusting the event's (possibly stale) target, so Bevy's per-command `world.flush()` (`bevy_ecs` 0.18.1, `command_queue.rs:185-189`) preserves the single-holder invariant across same-frame triggers — the guard is no longer needed.

**Affected:** GUI binary only (`src/action/*`, `src/input.rs`, `src/bootstrap.rs`, `src/main.rs`, and UI tests). The `ozmux_multiplexer` crate is untouched, and `src/multiplexer/commands.rs` continues to no-op these three actions, so there is no double-dispatch.

### Breaking Changes

Two marker-mutating session actions in a single frame (e.g. a very fast double `Cmd+R`) now **both** take effect — double `Cmd+R` creates two sessions instead of silently dropping the second — while the "exactly one `AttachedSession`" invariant is preserved. This is intended and is locked in by a new regression test (`two_new_session_events_same_frame_keep_marker_invariant`).

### Tests

Single-threaded suite is green (271 passed). Two anomalies are **pre-existing and out of scope** for this PR: an IME-suppression test (`dispatch_focused_key_suppressed_during_composition`, whose `make_app` harness omits the `.run_if(not(is_ime_composing))` guard) and a parallel-mode test-teardown SIGSEGV from real-PTY cleanup across threads. Both reproduce before this change; single-threaded mode is fully green.

### DIFF
--- a/src/action/session.rs
+++ b/src/action/session.rs
@@ -1,0 +1,388 @@
+//! Session-lifecycle shortcut actions dispatched as `EntityEvent`s.
+//!
+//! `NewSessionActionEvent` and `FocusSessionActionEvent` are triggered by
+//! the keyboard dispatcher (`crate::input::execute_action`) and handled by
+//! the observers below. The observers re-query the live `AttachedSession`
+//! marker rather than trusting the event's target, so two same-frame
+//! triggers preserve the single-holder invariant (Bevy flushes each
+//! observer's commands before the next queued trigger runs).
+
+use crate::multiplexer::{SessionCreatedAt, SessionNameCounter};
+use bevy::prelude::*;
+use ozmux_multiplexer::{AttachedSession, MultiplexerCommands, SessionMarker, SessionUiSubtree};
+
+/// Bevy Plugin that registers the session-action observers.
+pub struct OzmuxSessionActionPlugin;
+
+impl Plugin for OzmuxSessionActionPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_observer(on_new_session_action)
+            .add_observer(on_focus_session_action);
+    }
+}
+
+/// Request to mint a new session and attach it. Triggered by
+/// `ShortcutAction::NewSession`.
+#[derive(EntityEvent, Debug)]
+pub struct NewSessionActionEvent {
+    /// The session attached at dispatch time (trigger target only; the
+    /// observer re-queries the live marker).
+    #[event_target]
+    pub session: Entity,
+}
+
+/// Request to move session focus. Triggered by
+/// `ShortcutAction::FocusSession` and `ShortcutAction::FocusSessionNumber`.
+#[derive(EntityEvent, Debug)]
+pub struct FocusSessionActionEvent {
+    /// The session attached at dispatch time (trigger target only; the
+    /// observer re-queries the live marker).
+    #[event_target]
+    pub session: Entity,
+    /// Which session to focus.
+    pub target: FocusSessionTarget,
+}
+
+/// Selector for `FocusSessionActionEvent`, unifying `FocusSession{offset}`
+/// and `FocusSessionNumber{index}`. `Debug` is required because
+/// `FocusSessionActionEvent` derives `Debug`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum FocusSessionTarget {
+    Next,
+    Prev,
+    Last,
+    Number(u8),
+}
+
+/// Spawns a Session via `MultiplexerCommands` plus its UI subtree node,
+/// inserts `AttachedSession` + `SessionUiSubtree` + `SessionCreatedAt`
+/// on the session entity, and parents the subtree under the session.
+/// Returns the new session entity.
+///
+/// # Invariants
+///
+/// Inserts `AttachedSession` on the new session **without** removing it
+/// from any prior holder. A caller that must keep the "exactly one
+/// `AttachedSession`" invariant (the new-session path) is responsible for
+/// detaching the previous marker first; `bootstrap` calls it with no prior
+/// attached session, so it is safe there.
+pub(crate) fn spawn_attached_session(
+    commands: &mut Commands,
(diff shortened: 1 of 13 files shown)


src/action/session.rs
```diff
@@ -0,0 +1,430 @@
...
+    pub session: Entity,
+    /// Which session to focus.
+    pub(crate) target: FocusSessionTarget,
+}
+
+/// Selector for `FocusSessionActionEvent`, unifying `FocusSession{offset}`
+/// and `FocusSessionNumber{index}`. `Debug` is required because
+/// `FocusSessionActionEvent` derives `Debug`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum FocusSessionTarget {
```
not-elm: Change public struct

src/action/session.rs
```diff
@@ -0,0 +1,430 @@
...
+/// Request to move session focus. Triggered by
+/// `ShortcutAction::FocusSession` and `ShortcutAction::FocusSessionNumber`.
+#[derive(EntityEvent, Debug)]
+pub struct FocusSessionActionEvent {
+    /// The session attached at dispatch time (trigger target only; the
+    /// observer re-queries the live marker).
+    #[event_target]
+    pub session: Entity,
+    /// Which session to focus.
+    pub(crate) target: FocusSessionTarget,
```
not-elm: change public field

src/action/session.rs
```diff
@@ -0,0 +1,430 @@
...
+// applied first, those inserts would reference an entity that does not
+// exist yet and panic.
+fn on_new_session_action(
+    _trigger: On<NewSessionActionEvent>,
+    mut mux: MultiplexerCommands,
+    mut commands: Commands,
+    mut counter: ResMut<SessionNameCounter>,
+    attached_session: Query<Entity, (With<SessionMarker>, With<AttachedSession>)>,
+) {
+    dispatch_new_session(&mut commands, &mut mux, &mut counter, &attached_session);
```
not-elm: Please extract the logic.

src/action/session.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+fn on_focus_session_action(
+    trigger: On<FocusSessionActionEvent>,
+    mut commands: Commands,
+    sessions: Query<(Entity, Option<&SessionCreatedAt>), With<SessionMarker>>,
+    attached_session: Query<Entity, (With<SessionMarker>, With<AttachedSession>)>,
+) {
+    dispatch_focus_session(
```
not-elm: Please extract the logic.

not-elm (commented):
Plesae fix.

src/action/session.rs
not-elm: Done in 23e5680 — `NewSessionActionEvent` / `FocusSessionActionEvent` (plus `OzmuxSessionActionPlugin` and `OzmuxActionPlugin`) are now `pub(crate)`; they're only referenced within the binary crate.

src/action/session.rs
not-elm: Done in 23e5680 — the `#[event_target] session` fields are now `pub(crate)` (the `target` field already was).

src/action/session.rs
not-elm: Done in 23e5680 — inlined the logic into the observer (now `apply_new_session`) and removed the `dispatch_new_session` free function, matching the existing `apply_split` / `apply_swap_pane` style.

src/action/session.rs
not-elm: Done in 23e5680 — inlined into `apply_focus_session`; removed the `dispatch_focus_session` free function.

src/ui.rs
```diff
@@ -255,6 +255,7 @@ mod tests {
             .add_plugins(MultiplexerPlugin)
             .add_plugins(OzmuxConfigsPlugin)
             .add_plugins(OzmuxBootstrapPlugin)
+            .add_plugins(crate::action::OzmuxActionPlugin)
```
not-elm: Please explicitly import this at the top of the file.

not-elm (commented):
Plz fix.

src/ui.rs
not-elm: Done in cebfccf — added `use crate::action::OzmuxActionPlugin;` to the test module's imports and reference it by name in the builder (also applied the same in `ui/session.rs`).

PR #121, opened 2026-06-14
feat(ratatui-ozma): app-owned focus + passthrough input (host-tunnel)

not-elm:
## Problem

The `ratatui-ozma` SDK embeds CEF webviews as ratatui widgets. While a webview
held focus, the host routed every key to the page (CEF) and suppressed PTY
forwarding, so the ratatui app was blind to keystrokes — it could not run its
own focus navigation or react to app-level chords. There was no ergonomic way
for an app to (a) own which widget is focused and (b) still receive specific
keys while a webview is focused.

## Solution

Redesigns focus/input around a **host-tunnel (passthrough)** model: the app owns
focus entirely and receives its declared keys through plain
`crossterm::event::read`, with no SDK focus runtime.

- **Passthrough chords** are declared at register: `Webview::passthrough([KeyChord …])`.
  While that webview is focused, the host forwards a matching chord to the PTY
  (so the app reads it via `event::read`) **and** suppresses it from CEF (so it
  is not also typed into the page).
- **Focus is app-owned**, expressed at draw time via `WebviewWidget::focused(bool)`.
  `Ozma::flush` diffs the focused handle and emits the control-plane focus op,
  which the host applies to `bevy_cef`'s `FocusedWebview`.

Design / plan:
`docs/superpowers/specs/2026-06-15-ratatui-ozma-focus-input-redesign-design.md`,
`docs/superpowers/plans/2026-06-15-ratatui-ozma-focus-input-redesign.md`.

> Supersedes the earlier `FocusManager` / page-nav-glue approach prototyped on
> this branch — that surface (`FocusManager`, `NavKeymap`, `focusable`,
> `set_nav_keys`/`set_page_focus`, `Ozma::blur`) was removed in favor of
> passthrough, so the net diff against `main` is additive (nothing shipped to
> `main` is broken).

### Affected areas

- **SDK (`sdk/ratatui-ozma`)** — new `KeyChord` + `Webview::passthrough`;
  `WebviewWidget::focused` records the focused handle per frame; `Ozma::flush`
  emits the focus op on change; both examples rewritten to app-owned focus over
  plain `event::read`.
- **Host (`src/control_plane`, `src/input`, `src/inline_webview`, `src/webview_render`)** —
  the wire `passthrough` list is accepted and normalized once to
  `NormalizedChord` (bevy `KeyCode`) on `DynamicView` / a `PassthroughKeys`
  component; `fill_cef_keyboard_filter` mirrors the focused webview's chords into
  `bevy_cef`'s `CefKeyboardFilter`; `dispatch_focused_key` forwards a matching
  chord to the PTY, **hoisted above the global shortcut lookup** so a declared
  chord beats a shortcut; `sync_focused_webview` preserves app-declared focus
  against its per-frame re-derive.
- **Engine (`crates/ozma_tty_engine`, `src/input`)** — Alt/Meta → ESC encoding
  and macOS Option-glyph base-char recovery, so `Alt`-modified chords reach the
  PTY correctly.

### Data flow

```
webview focused + Alt+h pressed
  -> fill_cef_keyboard_filter: CefKeyboardFilter holds (entity, KeyH, Alt)  [CEF suppressed]
  -> dispatch_focused_key: passthrough_matches -> forward_to_active_terminal  [PTY]
  -> app reads Alt+h via crossterm event::read, moves its own focus
  -> WebviewWidget::focused(new) -> Ozma::flush -> Focus op -> FocusedWebview
```

### Dependency

Depends on **not-elm/bevy_cef#58** (the `passthrough` branch), which adds the
`CefKeyboardFilter` resource + public `KeyboardDeliverSet` used for CEF
suppression. The workspace tracks that branch via a git dependency; repoint to a
released tag once #58 merges.

### Testing

- SDK, engine, and host suites pass (host single-threaded:
  `cargo test -p ozmux-gui --bin ozmux-gui -- --test-threads=1`); `bevy_cef`
  `passthrough` branch suite green; `clippy --workspace` clean; `cargo fmt` +
  biome clean.
- Host coverage includes the precedence guarantee (`passthrough_wins_over_shortcut`),
  the CEF filter fill, and app-declared focus surviving the per-frame sync.
- Manual: `cargo run -p ratatui-ozma --example focus_grid` (2x2 grid, two
  webviews + two natives) — `Alt+h/j/k/l` moves focus and reaches `event::read`
  while a webview is focused, the same keys are suppressed from the page, bare
  keys type into the focused webview, and `q` quits only on a native cell.

### Notes / known limitations

- Passthrough key coverage is the documented set (letters, digits, `Tab`,
  `F1`–`F12`); keys outside it (arrows, `Enter`, `Esc`, …) are currently dropped
  silently at normalization — a follow-up should widen the table or surface a
  diagnostic.
- Focusing a webview on the same frame it first mounts can race the mount; start
  focus on a native widget (the examples do) or re-assert it on a later frame.


sdk/ratatui-ozma/examples/focus_grid.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        // On focus change, push the focus tint AND the chosen nav keymap to the
+        // webviews so the page glue intercepts the same chord this example uses.
+        let current = focused_id(focus);
+        if current != last_focus {
+            let keymap = keymap_payload(scheme);
+            let _ = ne.emit("focus", &(current == "ne"));
+            let _ = sw.emit("focus", &(current == "sw"));
+            let _ = ne.emit("__ozma.keys", &keymap);
+            let _ = sw.emit("__ozma.keys", &keymap);
```
not-elm: I’d like the SDK to provide dedicated wrapper methods for emitting focus and __ozma.keys.

not-elm: Done in 6b87c21. The SDK now provides typed wrappers instead of hand-rolled `emit`:

- New `NavKeymap` (single source for nav chords): presets `alt_hjkl()` / `ctrl_hjkl()` / `arrows()` + a builder, `Serialize` to the glue's `{mods,keys}` shape, and `match_key(&KeyEvent)` for native matching.
- `WebviewHandle::set_nav_keys(&NavKeymap)` wraps `emit("__ozma.keys")`; `WebviewHandle::set_page_focus(bool)` wraps `emit("__ozma.focus-state")` (the page observes it via `window.ozmux.on('__ozma.focus-state', cb)`).

The example now uses one `NavKeymap` for both native `match_key` and the glue push (`set_nav_keys`), so there are no raw reserved-string emits and the two can't drift. `FocusManager` stays pure (`navigate -> FocusSync`, I/O at the edge).

sdk/ratatui-ozma/examples/focus_grid.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/// The id of the currently-focused cell, or an empty string when none is.
+fn focused_id(focus: &FocusManager) -> String {
+    CELL_IDS
+        .into_iter()
+        .find(|id| focus.is_focused(id))
+        .unwrap_or("")
+        .to_owned()
+}
```
not-elm: I’d like to add a method to FocusManager that returns the currently focused ID.

not-elm: Done in cb98ea8. Added `FocusManager::focused_id(&self) -> Option<&str>` (a pure read-only accessor returning the currently-focused item's id, or `None`). The `focus_grid` example now uses `focus.focused_id()` directly and drops its `CELL_IDS` constant and the hand-rolled `focused_id` helper.

PR #136, opened 2026-06-16
feat(ui): tmux powerline window tabs + window-lifecycle fixes

not-elm:
## Problem

ozmux's bottom status bar did not show tmux windows as tabs — under the tmux control-mode backend there was no way to see or switch between windows from the GUI. Building that tab bar (a tmux-powerline look, the requested target UX) also surfaced several gaps in the tmux window-lifecycle projection that left the bar/render out of sync with tmux: windows wouldn't switch, newly created windows rendered black, killing a window left a stale tab, and the default `confirm-before` kill binding hung the app.

### Powerline window tabs (feature)

A tmux-powerline status bar at the bottom of the window: a slate session block, inactive windows as quiet `index:name` text, and the active window as an accent "chevron pill" (Nerd-Font powerline glyphs `` / ``). Per-window flags (zoom `Z`, bell `!`, activity `#`, silence `~`, marked `M`) are projected from tmux's `#{window_raw_flags}` — seeded via `list-windows` and kept live via a `refresh-client -B` subscription. Clicking a tab runs `select-window`.

- **`tmux_control_parser`** — `%subscription-changed` tolerates the `-` sentinel (`Option` fields) so the `@*` window subscription parses.
- **`ozmux_tmux` (`crates/tmux_session`)** — `WindowFlags` component + `#{window_raw_flags}` parser; seed + subscription plumbing; `TmuxWindowFlagsChanged` event/observer (conditional write).
- **`ozmux-gui` (`src/ui`, `src/font`, `src/theme`)** — the bar rebuild + pure color/flag-suffix helpers, a dedicated `PowerlineFont` handle to the bundled Nerd Font (override-proof), and theme constants.

### tmux window-lifecycle fixes (surfaced by the feature)

- **Bar never mounted** — `spawn_window_bar` raced `UiRoot` creation in `Startup` (no sync point between the two plugins' systems); moved to `PostStartup`.
- **Window switching** (next/previous/select-window) — tmux signals these with **only** `%session-window-changed` (no `%window-pane-changed`); the drain now re-queries the active pane so `ActiveWindow`/`ActivePane` move (verified against tmux 3.6a).
- **New windows rendered black** — `new-window` emits no `%layout-change`, so the pane was never projected; the drain re-enumerates `list-windows` on `%window-add` to fetch the layout.
- **kill-window left a stale tab** — handle `%unlinked-window-close` (what tmux actually emits) like `%window-close`, and order the bar rebuild `.after(TmuxProjectionSet)` so the despawn's `RemovedComponents` are seen.
- **kill-window hung the app** — the default binding is `confirm-before … kill-window`, a tmux client-side modal a `-CC` client cannot answer; ozmux now shows its own y/n confirm prompt and runs the inner command on confirm (iTerm2-style).

### Notes

- Each step was test-driven (parser, projection, pure UI functions, the confirm parser) and reviewed (per-task + a max-effort multi-agent pass). The branch passes `cargo test --workspace`, `cargo clippy --workspace`, and `cargo fmt --check`.
- UI change — screenshots recommended:
  <details><summary>Screenshots</summary>

  _Add a screenshot of the powerline window bar here (slate session block · accent chevron pill on the active window · flag suffixes)._

  </details>


crates/tmux_session/src/components.rs
```diff
@@ -31,6 +31,42 @@ pub struct TmuxWindow {
     pub name: String,
 }
 
+/// tmux per-window status flags, projected from `#{window_raw_flags}`.
+#[derive(Component, Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct WindowFlags {
```
not-elm: Use bitflags

not-elm (commented):
Please fix

src/ui/confirm_prompt.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+
+/// The active `confirm-before` prompt. Present while awaiting y/n; owns the
+/// keyboard like the copy-mode prompt and the session picker.
+#[derive(Resource, Default)]
+pub(crate) struct ConfirmPrompt {
+    /// The pending confirmation, if open.
+    pub(crate) open: Option<ConfirmState>,
+}
```
not-elm: Can we remove ConfirmPrompt and make ConfirmState a Resource?
For example, the run_if for handle_confirm_input can be rewritten as run_if(resource_exists::<ConfirmState>).
Also, I think we can split sync_confirm_ui into two separate systems for controlling visibility: hide_confirm_ui.run_if(resource_removed::<ConfirmState>) and visible_confirm_ui.run_if(resource_exists_and_changed::<ConfirmState>).

PR #148, opened 2026-06-18
feat(tmux): rename-window / rename-session via client-side command-prompt interception

not-elm:
## Problem

Under `tmux -CC` control mode, tmux's default rename bindings silently do nothing, because a `-CC` control client cannot render tmux's `command-prompt` (the same limitation that makes `confirm-before` unrenderable):

- `prefix + ,` → `command-prompt -I "#W" { rename-window -- "%%" }`
- `prefix + $` → `command-prompt -I "#S" { rename-session -- "%%" }`

So renaming a window or session from inside ozmux was impossible.

## Solution

Intercept the `command-prompt`-wrapped rename binding in `forward_keys_to_tmux` — the same pattern ozmux already uses for `confirm-before` and the copy-mode `command-prompt` — render an ozmux-owned bottom prompt bar prefilled with the current name, and on submit send a freshly-rebuilt, safely-quoted rename command with an explicit target id. Both **rename-window** (`prefix + ,`) and **rename-session** (`prefix + $`) are handled. Affected: `ozmux_tmux` (`crates/tmux_session`) and the GUI binary (`src/`).

- **Recognizer** (`parse_command_prompt_rename`): classifies a binding by *shape*, not substring — first token `command-prompt`, inner verb `rename-window`/`renamew`/`rename-session`/`rename`, canonical `%%`/`%1` placeholder. Rejects multi-input prompts by substitution **arity** (`%2`–`%9`), which is robust to the `-l` literal flag and commas inside a single quoted prompt; ignores decorated templates and `run-shell`-style substring matches.
- **Command builders** `rename_window_command` / `rename_session_command` use the existing `quote()` helper and an explicit `-t @id` / `-t $id` captured at prompt-open — quote-safe (handles spaces/quotes/empty) and race-safe vs. tmux's fragile `%%` substitution.
- **New `RenamePromptPlugin`** (`src/ui/rename_prompt.rs`): a bottom prompt bar modeled on `confirm_prompt.rs` / `copy_search.rs`. While open it owns the keyboard **and** the mouse wheel; an armed-skip drains the opening key; Enter submits, Esc cancels.
- **16-param ceiling**: the new resource/query access is folded into `forward_keys_to_tmux`'s existing param tuple via a `RenameParams` `SystemParam` bundle.
- **Session-label display**: a real `tmux 3.6a` `-CC` trace showed `rename-session` emits only `%session-renamed` (no follow-up `%session-changed`), which `event_pump` did not map — added `ControlEvent::SessionRenamed → TmuxSessionChanged` so the window bar's session name refreshes.

### Testing

- **Unit**: recognizer matrix (default window/session, `renamew`/`rename` aliases, brace vs. quoted templates, `%1`, multi-input rejection, `-l` + comma, comma-in-quoted-prompt, decorated / copy-mode-search / `confirm-before` / `run-shell` rejection); input-step + command-builder tests; `%session-renamed` routing test.
- **Suites green**: `ozmux_tmux` 132, GUI bin 352. `cargo fmt` / `clippy` clean for the changed files.
- **Pending**: manual macOS GUI smoke (`prefix + ,` / `prefix + $` open the bar, edit + Enter renames, Esc cancels, keys don't leak to the pane). <details><summary>Screenshots</summary>Pending manual GUI smoke.</details>

### Note

`rename_prompt.rs` keeps its own brace-aware `tokenize`, deliberately separate from `confirm_prompt.rs`'s (different contracts: rename must treat `{`/`}` as tokens, confirm must not).


src/ui/rename_prompt.rs
```diff
@@ -0,0 +1,638 @@
...
+/// and non-`command-prompt` commands).
+///
+/// Recognition is by command shape, not substring: the first token must be
+/// `command-prompt`; the inner template (a `{ ... }` brace body or a quoted
+/// template argument) must be exactly `<rename-verb> [--] <placeholder>` where
+/// the verb is `rename-window`/`renamew`/`rename-session`/`rename` and the
+/// placeholder is `%%` or `%1`. Multi-input prompts are rejected by arity (any
+/// `%2`..`%9` reference), which is robust against the `-l` literal flag and
+/// commas embedded in a single quoted prompt.
+pub(crate) fn parse_command_prompt_rename(command: &str) -> Option<RenameKind> {
```
not-elm: Convert this into a method:
```rust
impl RenameKind{
	pub fn parse(command: &str) -> Option<Self>{
		...
	}
}
```

src/ui/rename_prompt.rs
```diff
@@ -0,0 +1,638 @@
...
+#[derive(Debug, PartialEq, Eq)]
+enum RenameStep {
+    Continue,
+    Submit,
+    Cancel,
+}
+
+/// Applies one key to the prompt: Enter submits, Escape cancels, Backspace
+/// deletes the last char, a character is appended. Other keys are ignored.
+fn apply_rename_key(prompt: &mut RenamePrompt, key: &Key) -> RenameStep {
```
not-elm: Convert this into a method.

not-elm (commented):
Please fix

not-elm:
Addressed in bd288ab — converted the type-associated free functions into methods:

- `parse_command_prompt_rename(command)` → `RenameKind::parse(command)`
- `apply_rename_key(&mut prompt, key)` → `RenamePrompt::apply_key(&mut self, key)`

I also applied the same conversion to the two analogous siblings so the result is uniform rather than half-converted:

- `submit_command(&prompt)` → `RenamePrompt::submit_command(&self)`
- `rename_label(&subject)` → `RenameSubject::label(&self)`

All call sites and tests updated; build/fmt/clippy clean, 22 rename tests green. Let me know if you'd prefer I leave `submit_command`/`label` as free functions instead.

PR #158, opened 2026-06-19
feat(ozma): Ozma mode — single-terminal PTY with tmux picker and mode lifecycle

not-elm:
## Problem

ozmux lacked a single-terminal mode for users who don't need tmux multiplexing. Additionally, the open-picker shortcut in Ozma mode was silently broken: pressing it set `picker.open = true` but nothing could act on it — the picker UI never appeared and the terminal subsequently lost keyboard input entirely (because `forward_keys_to_ozma` kept clearing the event queue while `picker.open` was stuck `true`).

### New `AppMode::Ozma` state

Adds an `ozma_mode` crate that implements a single-PTY terminal filling the window. When active, the app runs one shell process directly — no tmux multiplexer involved. Key pieces:

- `OzmaModePlugin` — spawns the terminal entity on `OnEnter(AppMode::Ozma)`, tears it down on `OnExit`, and runs a resize system that keeps the PTY grid filling the window each frame.
- `OzmaInputPlugin` (`src/ozma_input.rs`) — keyboard and paste forwarding in Ozma mode. Intercepts `ShortcutAction::OpenPicker` to show the session picker, `ShortcutAction::DetachSession` is a no-op (no session to detach from), and all other keys are forwarded to the PTY.
- IME support: `ime_policy_system` extended to recognise `OzmaTerminal` and enable IME for the Ozma terminal pane.

### Mode lifecycle

- `StartupMode` config enum (`Ozma | Ozmux | AutoAttach`) added to `~/.config/ozmux/config.toml` under `[startup]`.
- Boot always starts in `AppMode::Ozmux` so `OnEnter(AppMode::Ozmux)` Startup deferred commands (e.g. `init_atlas_image`) are flushed before any `OnEnter(AppMode::Ozma)` fires. `on_enter_ozmux_picker` handles `StartupMode::Ozma` by immediately scheduling a transition to `AppMode::Ozma`.
- `on_exit_ozmux` sends `detach-client` to tmux, drops the connection, removes `TmuxPresence`, and triggers `TmuxConnectionReset`. `on_tmux_connection_closed` observer transitions back to `AppMode::Ozma`.
- `DetachSession` shortcut in Ozmux mode triggers `on_exit_ozmux` gracefully.
- All tmux UI `Update` systems gated via `OzmuxActiveSet` (`.run_if(in_state(AppMode::Ozmux))`); dialog backdrop hidden on Ozmux exit.

### Picker fix

`handle_picker_input`, `refresh_picker_on_open`, and `sync_picker_ui` were all gated to `AppMode::Ozmux`. Removed those gates so the picker operates in both modes. `refresh_session_ozmux_sock` stays gated as it is tmux-specific.

### Code-review fixes

- `ime_policy_system` (`src/input/ime.rs`): moved `mut primary_window` to first parameter position, satisfying the mutable-params-first convention.
- `drain_tmux_events` (`crates/tmux_session/src/plugin.rs`): removed the three redundant direct resource clears (`*enumeration`, `keybindings.clear()`, `copy_queries.clear()`) from the `Closed` branch — `on_connection_reset` already resets them when `TmuxConnectionReset` fires.

### DIFF
--- a/src/tmux.rs
+++ b/src/tmux.rs
@@ -18,7 +18,9 @@ use divider_handle::DividerHandlePlugin;
 use input::InputPlugin;
 use mouse::MousePlugin;
 use ozma_mode::AppMode;
-use ozmux_tmux::{TmuxConnection, TmuxConnectionReset, TmuxPresence, TmuxSessionPlugin};
+use ozmux_tmux::{
+    TmuxConnection, TmuxConnectionClosed, TmuxConnectionReset, TmuxPresence, TmuxSessionPlugin,
+};
 use pane_focus::PaneFocusPlugin;
 use render::RenderPlugin;
 use window_bar::WindowBarPlugin;
@@ -35,6 +37,7 @@ impl Plugin for OzmuxTmuxPlugin {
         app.configure_sets(Update, OzmuxActiveSet.run_if(in_state(AppMode::Ozmux)));
         app.add_systems(OnEnter(AppMode::Ozmux), on_enter_ozmux);
         app.add_systems(OnExit(AppMode::Ozmux), on_exit_ozmux);
+        app.add_observer(on_tmux_connection_closed);
         app.add_plugins((
             TmuxSessionPlugin,
             RenderPlugin,
@@ -47,6 +50,13 @@ impl Plugin for OzmuxTmuxPlugin {
             PaneFocusPlugin,
         ));
     }
+}
+
+fn on_tmux_connection_closed(
+    _ev: On<TmuxConnectionClosed>,
+    mut next_mode: ResMut<NextState<AppMode>>,
+) {
+    next_mode.set(AppMode::Ozma);
 }
 
 fn on_enter_ozmux(mut commands: Commands) {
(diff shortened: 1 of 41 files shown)


src/tmux.rs
```diff
@@ -17,16 +17,27 @@ use dialog::DialogPlugin;
...
+
 /// Bevy plugin aggregating all tmux runtime sub-plugins.
 pub struct OzmuxTmuxPlugin;
 
 impl Plugin for OzmuxTmuxPlugin {
     fn build(&self, app: &mut App) {
+        app.configure_sets(Update, OzmuxActiveSet.run_if(in_state(AppMode::Ozmux)));
+        app.add_systems(OnEnter(AppMode::Ozmux), on_enter_ozmux);
+        app.add_systems(OnExit(AppMode::Ozmux), on_exit_ozmux);
+        app.add_observer(on_tmux_connection_closed);
```
not-elm: `app.add_systems`などはできる限りメソッドチェーンで宣言するようにしてください。
ルールファイルにも記載してください。

PR #188, opened 2026-06-22
fix: set a fallback TERM at startup so bundled launches keep line editing

not-elm:
## Problem

Running ozmux from the bundled `.app` (Finder/`open`), **Backspace stops deleting** — the character stays and the cursor advances — while `cargo run` (debug) is fine.

## Root cause

A Finder-launched `.app` inherits launchd's minimal environment where **`TERM` is empty**. ozmux spawns `tmux -CC` (and the native ozma PTY) with that environment, so the pane shells come up with an **empty `$TERM`**. With no `TERM`, zsh can't load terminfo and its line editor (ZLE) silently degrades — Backspace included — even though `bindkey '^?'` (`backward-delete-char`), `stty erase` (`^?`), and the delivered `0x7f` byte are all correct. `cargo run` only worked because it inherits a real `TERM` from the launching terminal.

Investigation ruled out every other layer with evidence: winit→ozmux→tmux key encoding (`send-keys -- BSpace` → `0x7f`), tmux key translation, `stty erase`, the `--no-default-features`/`dist` build difference (only the CEF `debug` feature), and IME (separate pre-existing issue). A pane forced to an empty `$TERM` reproduces the broken Backspace exactly.

## Fix

Mirror [`alacritty_terminal::tty::setup_env`](https://github.com/alacritty/alacritty): at the top of `main()` (before any task-pool thread spawns), fill `TERM=xterm-256color` + `COLORTERM=truecolor` **only when the inherited `TERM` is unset or empty**.

- `xterm-256color` is exactly Alacritty's fallback when the `alacritty` terminfo is absent (it is, on stock macOS), and is a system-standard entry in `/usr/share/terminfo`.
- A usable inherited `TERM` is left untouched → terminal launches unchanged.
- Process-wide placement covers both the tmux panes and the native ozma PTY.

This direction was cross-checked against Alacritty's source via parallel research (Codex + agent), both High confidence.

## Verification

- `cargo test --bin ozmux` → 16 passed (incl. 3 new `term_fallback` tests)
- `cargo clippy --bin ozmux` → no warnings
- Rebuilt the `dist` bundle and confirmed Backspace deletes (and `$TERM` is non-empty) in a Finder-launched pane.

https://claude.ai/code/session_01KuPjM8h6GDuJHiSF9pSoJ5

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,22 +42,9 @@ use ui::{
 };
 
 fn main() {
-    // Mirror alacritty_terminal::tty::setup_env: a child shell (a tmux pane or
-    // the native ozma PTY) whose TERM is empty/unset cannot load terminfo, so
-    // zsh's line editor (ZLE) — Backspace included — silently breaks. A bundled
-    // .app launched from Finder inherits launchd's empty TERM, so fill a portable
-    // default before any PTY child is spawned. `xterm-256color` is exactly what
-    // Alacritty falls back to when the `alacritty` terminfo is absent (it is, on
-    // stock macOS); `COLORTERM` advertises the 24-bit color that entry omits.
-    if let Some(term) = term_fallback(std::env::var("TERM").ok().as_deref()) {
-        // SAFETY: this runs at the very top of main(), before App::new() spawns
-        // any task-pool threads, so no other thread can read the environment
-        // concurrently with these writes.
-        unsafe {
-            std::env::set_var("TERM", term);
-            std::env::set_var("COLORTERM", "truecolor");
-        }
-    }
+    // NOTE: must run before App::new() spawns any thread — it writes process
+    // env vars, which is unsound once other threads may read the environment.
+    ensure_terminfo_env();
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -42,6 +42,23 @@ use ui::{
...
+    if let Some(term) = term_fallback(std::env::var("TERM").ok().as_deref()) {
+        // SAFETY: this runs at the very top of main(), before App::new() spawns
+        // any task-pool threads, so no other thread can read the environment
+        // concurrently with these writes.
+        unsafe {
+            std::env::set_var("TERM", term);
+            std::env::set_var("COLORTERM", "truecolor");
+        }
+    }
+
```
not-elm: extract into private function .

not-elm: Done in cf31758 — extracted into a private `ensure_terminfo_env()`; `main()` now just calls it. The before-threads invariant is preserved and documented under `# Invariants`, with the SAFETY note on the unsafe env write. Tests + clippy green.

PR #201, opened 2026-06-25
fix(tmux): make $OZMA_SOCK resolvable in the adopted tmux pane

not-elm:
## Problem

In tmux mode (in-place `tmux -CC` adoption), `$OZMA_SOCK` was unreachable in the adopted/first pane: `echo $OZMA_SOCK` is empty, and — more importantly — the `ratatui-ozma` SDK could not resolve the control socket there.

Root cause (verified empirically on tmux 3.6b):

- ozmux only injects `$OZMA_SOCK` *after* it adopts the `tmux -CC` connection, which is strictly after the first pane's shell has already forked. A running process's environment can't be changed from outside, so that pane never sees it.
- The injection used `set-environment -g` (global scope). The SDK is designed to recover the value from any pane via `tmux show-environment OZMA_SOCK`, but a **global-scope** variable is invisible to **session-scope** `show-environment` (returns `unknown variable`, exit 1). So the documented recovery path silently failed in the adopted pane. The existing tests passed only because the test harness set the var at session scope while production used global scope.

## Solution

Propagate `$OZMA_SOCK` at **session scope** in addition to the global set, in `src/tmux/webview_tokens.rs` (the tmux webview-tokens plugin):

- New `bind_ozma_sock_to_session` system sends `set-environment -t '$<id>' OZMA_SOCK <path>` for the attached session. tmux injects `$TMUX` into every pane, so the SDK's `tmux show-environment OZMA_SOCK` now recovers the socket from any pane — including the first.
- Gated on `Changed<TmuxSession>` (not `Added`), matching how `window_title`/`window_bar` already react to session switches — so it fires on the initial adopted session and re-fires whenever the control client switches to a different session (`on_session_changed` reuses one session entity, so a switch is a re-insert, not an `Added` edge).
- The existing global `-g` set is kept: it seeds the process environment of panes spawned *after* it, which is what makes `echo $OZMA_SOCK` work in newly-created panes.

Verified end-to-end against a real tmux server (session-scope set → `tmux show-environment OZMA_SOCK` in the first pane returns the path, exit 0). New unit tests cover the initial set, the switch re-send, and the no-control-plane no-op; full binary suite (334 tests) passes, clippy/fmt clean.

> Note: a literal `echo $OZMA_SOCK` in the adopted first pane stays empty by design — a running process's environment cannot be changed from outside. SDK-based programs are the supported path there and now work via `tmux show-environment`. Newly-created panes get it in their process env as before.

### DIFF
--- a/src/tmux/webview_tokens.rs
+++ b/src/tmux/webview_tokens.rs
@@ -10,7 +10,7 @@
 //!
 //! - A global `-g` set (on the attach edge) so the process environment of any
 //!   pane spawned *after* the set inherits `$OZMA_SOCK` directly.
-//! - A per-session `-t` set (on each newly projected session) so
+//! - A per-session `-t` set (on each session add or switch) so
 //!   `tmux show-environment OZMA_SOCK` recovers the value from *any* pane of the
 //!   session — including a pane that forked before the global set (the adopted
 //!   control-mode pane, and any pane attached to a pre-existing session). A
@@ -38,7 +38,10 @@ impl Plugin for WebviewTokensPlugin {
             bind_tmux_pane_tokens.run_if(any_with_component::<TmuxClient>),
         )
         .add_systems(Update, refresh_ozma_sock.run_if(tmux_client_added))
-        .add_systems(Update, bind_ozma_sock_to_session.run_if(tmux_session_added))
+        .add_systems(
+            Update,
+            bind_ozma_sock_to_session.run_if(tmux_session_changed),
+        )
         .add_systems(Last, cleanup_ozma_sock.run_if(on_message::<AppExit>));
     }
 }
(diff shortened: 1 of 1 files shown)


src/tmux/webview_tokens.rs
```diff
@@ -74,6 +95,44 @@ fn refresh_ozma_sock(mut client: TmuxClientMut<'_, '_>, control: Option<Res<Cont
...
+/// every pane, so the ratatui-ozma SDK recovers the value with
+/// `tmux show-environment OZMA_SOCK` — but only when it is set at session scope,
+/// which this system does. The session is targeted by id (`$N`), which is stable
+/// and unique, rather than name, which may be empty or duplicated.
+///
+/// No-op when the control plane is absent (no `ControlPlaneHandle`).
+fn bind_ozma_sock_to_session(
+    mut client: TmuxClientMut<'_, '_>,
+    sessions: Query<&TmuxSession, Changed<TmuxSession>>,
+    control: Option<Res<ControlPlaneHandle>>,
```
not-elm: `run_if(resource_exists::<ControlPlaneHandle>)`を使って事前に存在チェックすればOptionにする必要がなくなるはずです

PR #204, opened 2026-06-25
fix(tmux): propagate UTF-8 LC_CTYPE to panes so typed CJK isn't garbled

not-elm:
## Problem

Typing Japanese (or any CJK) via the IME into a tmux pane and "holding"/typing continuously prints garbage like `ÔøΩ<0083><0080>` instead of the characters:

```
% dadadadaÔøΩ<0083><0086>ÔøΩ<0082>ÔøΩÔøΩ<0083><0088>ÔøΩ<0083><0080>…
```

Decoded, each `ダ` (`E3 83 80`) arrives as the codepoints **U+FFFD** (`ÔøΩ`) + `<0083>` + `<0080>` — the UTF-8 lead byte `E3` orphaned, the continuation bytes surfacing as placeholders.

Investigation (see notes below) proved ozmux is **not** the corruptor:

- A pane `xxd` capture showed ozmux **sends clean** `e3 83 80 …`.
- A unit test through the real VT parser shows the renderer only emits `�`+C1 when a wide char is split by an escape — i.e. the *echo stream itself* is malformed.
- `tmux pipe-pane` confirmed: zsh's line editor (ZLE) in the **C/POSIX locale** re-renders each typed wide char byte-by-byte (orphaned lead byte + `<0083>`-style `nicechar` placeholders). The user confirmed `LC_CTYPE="C"` in the pane.

So the root cause is that the pane's shell runs in a non-UTF-8 locale. It only affects **typed** CJK (ZLE echo) — program output of UTF-8 passes straight through.

Why the existing fix didn't cover it: ozmux **adopts the user's own `tmux -CC`** rather than spawning tmux, so `ensure_utf8_locale_env` (which only normalizes ozmux's own process env) never reaches a pre-existing / stripped-env server's panes. Same locale family as #202, different symptom.

## Solution

On the tmux attach edge, push `set-environment -g LC_CTYPE <utf8>` into the adopted server's global environment so panes spawned afterward inherit a UTF-8 locale — mirroring the `$OZMA_SOCK` global propagation in `src/tmux/webview_tokens.rs`.

Affected (engine/tmux integration only):

- `src/tmux/locale.rs` *(new)* — `TmuxLocalePlugin`: a `propagate_utf8_ctype` system gated on the `Added<TmuxClient>` attach edge sends the `set-environment -g LC_CTYPE` command.
- `src/main.rs` — `utf8_ctype_for_panes()` resolves the value (first UTF-8 of `LC_ALL` / `LC_CTYPE` / `LANG`, else `en_US.UTF-8`); `ensure_utf8_locale_env` already guarantees a UTF-8 result on macOS.
- `src/tmux.rs` — registers `TmuxLocalePlugin`.

### Scope / limitations (documented in `src/tmux/locale.rs`)

- `set-environment -g` reaches panes spawned **after** the set; an already-running shell (incl. the current one) must be restarted, or `export LC_CTYPE=en_US.UTF-8`.
- Only `LC_CTYPE` is set. A non-UTF-8 `LC_ALL` deliberately exported into the server's env outranks it and is intentionally left untouched (clearing it would clobber a deliberate UTF-8 `LC_ALL`).

### Tests

`resolve_utf8_ctype` precedence/fallback is unit-tested; the attach-edge system is tested for "sends a global UTF-8 `LC_CTYPE`" and "runs once per attach, not every frame". Full `ozmux` + `ozmux_tmux` suites pass (486 tests); clippy + fmt clean.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -222,6 +222,48 @@ fn is_utf8_locale(value: &str) -> bool {
     upper.contains("UTF-8") || upper.contains("UTF8")
 }
 
+/// The UTF-8 `LC_CTYPE` ozmux advertises to tmux panes spawned after attach
+/// (see `crate::tmux::locale`), so their shells render typed wide characters
+/// correctly instead of as `<00xx>` placeholders.
+///
+/// A pane shell stuck in the C/POSIX locale makes zsh's line editor re-render
+/// every typed multibyte glyph as its per-byte C-locale notation (an orphaned
+/// lead byte plus `<0083>`-style placeholders), which ozmux then faithfully
+/// displays. This happens whenever the tmux server ozmux adopts was itself
+/// started without a UTF-8 locale (e.g. a pre-existing server, or a launchd
+/// `.app` with stripped env). Pushing this value into tmux's environment fixes
+/// newly-spawned panes.
+///
+/// Resolves to the first UTF-8 locale among `LC_ALL` / `LC_CTYPE` / `LANG`,
+/// skipping any non-UTF-8 entries, falling back to [`UTF8_CTYPE_FALLBACK`].
+///
+/// This deliberately differs from [`utf8_locale_fallback`]: that one honors
+/// tmux's first-non-empty-wins precedence to decide *whether* ozmux's own
+/// process needs the fallback, so a leading non-UTF-8 value (e.g. `LC_ALL=C`)
+/// makes it return the fallback. Here we instead want a concrete UTF-8 *value*
+/// to advertise to panes, so a non-UTF-8 entry is skipped rather than allowed to
+/// win. Because `ensure_utf8_locale_env` runs first, on macOS this always
+/// resolves to a UTF-8 value even when the inherited environment selected the
+/// C/POSIX locale.
+pub(crate) fn utf8_ctype_for_panes() -> String {
+    resolve_utf8_ctype(
+        std::env::var("LC_ALL").ok().as_deref(),
+        std::env::var("LC_CTYPE").ok().as_deref(),
+        std::env::var("LANG").ok().as_deref(),
+    )
+}
+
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -222,6 +222,48 @@ fn is_utf8_locale(value: &str) -> bool {
...
+///
+/// This deliberately differs from [`utf8_locale_fallback`]: that one honors
+/// tmux's first-non-empty-wins precedence to decide *whether* ozmux's own
+/// process needs the fallback, so a leading non-UTF-8 value (e.g. `LC_ALL=C`)
+/// makes it return the fallback. Here we instead want a concrete UTF-8 *value*
+/// to advertise to panes, so a non-UTF-8 entry is skipped rather than allowed to
+/// win. Because `ensure_utf8_locale_env` runs first, on macOS this always
+/// resolves to a UTF-8 value even when the inherited environment selected the
+/// C/POSIX locale.
+pub(crate) fn utf8_ctype_for_panes() -> String {
```
not-elm: src/tmux/locale.rsに移動してほしい

not-elm: `src/tmux/locale.rs` に移動しました（a980936）。

- `utf8_ctype_for_panes` と純粋ヘルパ `resolve_utf8_ctype`（＋ユニットテスト）を、唯一の呼び出し元 `propagate_utf8_ctype` の隣へ移動。モジュール外から使われなくなったため `private` にしています。
- 共有プリミティブ `is_utf8_locale` / `UTF8_CTYPE_FALLBACK` は起動時の `ensure_utf8_locale_env` / `utf8_locale_fallback` でも使うため `src/main.rs` に残し、`pub(crate)` で公開しました。

挙動の変更はありません。テスト（486件）・clippy・fmt・doc いずれもクリーンです。

PR #229, opened 2026-07-01
feat(shortcuts): tmux-style leader keys — flat `<Leader>` bindings + modifier-tap leaders

not-elm:
## Problem

ozmux's shortcut layer only allowed a single direct chord per action. This adds tmux-style **leader (prefix) key** support, so an action can fire as `leader → key` (e.g. `Ctrl+A` then `d`), independent of tmux's own prefix.

## Solution — flat `<Leader>` bindings

The shortcut config is a **single flat `[shortcuts]` table**. `leader` sets the leader; each action's value is a direct chord (`"Cmd+V"`), a leader-scoped chord (`"<Leader>s"` = press the leader, then `s`), or `""` (unbind). Direct vs leader-scoped is encoded by the `<Leader>` token in the value — there is no separate bindings sub-table.

```toml
[shortcuts]
leader = "Ctrl+A"
paste           = "Cmd+V"
enter-copy-mode = "<Leader>s"   # Ctrl+A then s
detach-session  = "<Leader>d"   # Ctrl+A then d
```

- **Config** (`crates/ozmux_configs`): one serde-declared `Shortcuts` struct (each action `Option<Binding>`, `Binding = Direct | Leader`); the previous parallel `Bindings` / `PrefixBindings` schema (a two-table design that defined the same action set twice) is gone. Load-time validation (fatal `exit 2`): duplicate direct chords; duplicate leader chords; a `<Leader>` binding with no leader; a leader that shadows a direct chord; an unmappable leader key. A direct `s` and a `<Leader>s` never collide (different key-spaces).
- **Runtime** (`src/input`): named-field `Shortcuts` resource (`direct` / `prefix` / `leader`) plus a pure, unit-tested `step_leader` state machine, wired into both the tmux and Default keyboard dispatchers through a shared `LeaderPending`.
- **Default-mode PTY gating** (`src/input/keyboard.rs`): the leader chord is reserved and `dispatch_input` suppresses the pending second key, so leader sequences never leak into the terminal.
- **Docs**: `docs/configs.md` documents the flat `[shortcuts]` format and the per-mode action asymmetry (`enter-copy-mode` is Default-mode only; `detach-session` is tmux only).

### Compatibility (breaking)

Not backward compatible. The retired `[shortcuts.bindings]` / `[shortcuts.prefix_bindings]` tables and the old `prefix` key are unknown fields under `deny_unknown_fields`, so a legacy config is a TOML parse error. The host treats parse errors as **warn + fall back to defaults** (only validation errors `exit 2`), so a stale config silently reverts the *entire* config to defaults with a warning — migrate to the flat `<Leader>` format.

## Modifier-tap leaders

Building on the flat design above, `leader` may now also be a **bare modifier to tap** — not just a chord. `leader = "Cmd"` fires the leader on a Cmd **tap** (press+release with no intervening key or mouse press, within a timeout), so `paste = "<Leader>p"` means "tap Cmd, then p". Chord leaders (`Ctrl+A`) still work; the two forms coexist.

```toml
[shortcuts]
leader = "Cmd"                 # tap the bare Cmd modifier — or a chord like "Ctrl+A"
leader-tap-timeout-ms = 300    # tap window (ms); 0 reverts to 300
paste = "<Leader>p"            # tap Cmd, then p
```

- **Config**: `leader` is now `Leader::{Chord | ModifierTap(Cmd/Ctrl/Alt)}` (`Shift` is excluded — too noisy as a tap). New kebab-case `leader-tap-timeout-ms` (default 300, 0 → 300). Tap leaders skip the shadow / unmappable validations (a tap is not a chord).
- **Runtime**: a mode-agnostic `detect_modifier_tap` system (new `LeaderGate::Detect` set, ordered `Detect → Read → Advance`) watches all `KeyboardInput` press **and** release (plus mouse presses) and, via a pure `step_tap` machine on `Res<Time<Real>>`, engages the shared `LeaderPending` on a valid tap. The second key is consumed by the **unchanged** `step_leader` pending arm — the chord-leader path is a no-op regression. `forward_keys_to_tmux` was added to `LeaderGate::Advance` to close a same-frame ordering gap for tmux.
- **Disarm / no false-fire**: any intervening non-target key-down **or any mouse press** cancels the tap (so `Cmd+Q` and `Cmd+click` never mis-fire); `detect_modifier_tap` runs every frame while a tap leader is configured so a mouse press disarms even on a keyboard-less frame. The in-progress tap is cleared on mode change, webview focus change, and window unfocus.

### Testing

All suites green: `ozmux_configs` (config + validation), the `ozmux` binary's `shortcuts` (incl. pure `step_tap` unit tests + **App-level `detect_modifier_tap` wiring tests**, notably mouse-disarm-on-a-keyboard-less-frame), `default_mode`, `keyboard`, and `ozmux_tmux` — **483 workspace tests, 0 failures**; `clippy` + `fmt` clean, 0 warnings. Hardened through two spec reviews (Codex + Claude Code Agent grounding claims against local bevy 0.18 / winit 0.30 source), per-task + whole-feature multi-agent reviews, a `/code-review max` pass, and a `/simplify` pass. A pre-existing non-hermetic config test (it read the developer's real `~/.config`) was made hermetic in passing.

**Not yet done — manual macOS GUI smoke test** (keyboard/modifier dispatch can't be validated headless; winit modifier `Released` delivery is the one premise not statically provable). Recommended before merge:
- **Chord leader** — `leader = "Ctrl+A"`, `detach-session = "<Leader>d"`: `Ctrl+A d` detaches; `Ctrl+A <unbound>` is a no-op; tmux `C-b` still works.
- **Tap leader** — `leader = "Cmd"`, `paste = "<Leader>p"`: tap Cmd then `p` pastes; `Cmd+Q` still quits (no mis-fire); `Cmd+click` does **not** engage the leader; long-holding Cmd (>300ms) then releasing does **not** engage; verify in both Default and tmux modes.


src/input/shortcuts.rs
```diff
@@ -5,28 +5,103 @@
...
+            .add_systems(OnExit(AppMode::Default), reset_leader_pending)
+            // NOTE: webview focus moves on mouse clicks (no `KeyboardInput`), so the
+            // keyboard dispatchers never see the round-trip; without this reset a
+            // leader engaged before a mouse-only webview focus/blur would consume the
+            // next terminal keystroke as its second key.
+            .add_systems(
+                Update,
+                reset_leader_pending
+                    .run_if(resource_exists_and_changed::<FocusedWebview>)
+                    .before(LeaderGate::Detect),
```
not-elm: Updateのシステムをタプルでグループ化して

PR #262, opened 2026-07-09
feat(font): resolve terminal + UI font by system family name

not-elm:
## Problem

The `[font]` config section only accepted absolute / `~`-prefixed TTF file paths
(`normal`, `bold`, `italic`, `bold_italic`). Users had to locate each face file
on disk and wire all four paths by hand — there was no way to select an
installed system font by its family name, nor to pull weight/style variants from
a single family.

## Solution

Repurpose the `[font]` keys to resolve **font-family names** against the system
font database (via `fontique`), keeping the symmetric
`normal` / `bold` / `italic` / `bold_italic` shape.

- **Config schema** (`orzma_configs`): the `[font]` keys `normal` / `bold` /
  `italic` / `bold_italic` change meaning from TTF file paths to system
  font-family names. `normal` is the base family; each per-face key, when
  omitted, derives from `normal`. Omitting all of them uses the bundled
  JetBrains Mono Nerd Font.
- **Resolver** (`src/font/resolve.rs`, new): a pure `fontique`-backed function
  that turns `(family, weight/style)` into `(font bytes, .ttc face index)`.
  Pure over a borrowed collection + source cache, so it is unit-testable with
  injected fonts instead of the host's installed fonts.
- **Renderer** (`orzma_tty_renderer`): `TerminalFonts::from_faces` loads each
  primary face at its own `.ttc` collection index and records the regular
  face's index, so later `ttf-parser` metric reparses (`cell_metrics_px`,
  `em_scale_of`) target the correct face. `from_bytes` now delegates with
  index 0 (signature unchanged for existing callers).
- **UI** (`src/`): `TerminalUiFont` becomes bevy_text's `FontSource` enum;
  `bridge_font_config` resolves each face and sets `FontSource::Family` when the
  regular face resolves from a system family, otherwise falling back to bundled
  bytes. An unknown family — or a present family missing a requested
  weight/style — falls back per the resolver's contract.
- **Build**: enable bevy's `system_font_discovery` feature.
- **Docs**: `docs/configs.md` updated to the family-name `[font]` schema and its
  fallback behavior.

Affected areas: engine renderer (`orzma_tty_renderer`), config crate
(`orzma_configs`), UI (`src/ui/*`), and docs. Tests cover per-face family
override parsing plus a deterministic bridge success-path test that registers
bundled JBM faces under known family names in the test app's `FontCx` before
`Startup`, proving a configured family resolves through both `TerminalFonts` and
`TerminalUiFont` and that `bold.or(normal)` is honored on each side.

### Breaking Changes

The `[font]` keys `normal` / `bold` / `italic` / `bold_italic` change meaning
from **TTF file paths** to **system font-family names**. The key names are
unchanged, so existing configs still parse — but a value that used to be a file
path (e.g. `normal = "/Users/me/MyMono.ttf"`) is now read as a family name,
fails to resolve, and silently falls back to the bundled JetBrains Mono. Users
who set custom font-file paths must replace them with the family name of an
installed font (e.g. `normal = "JetBrains Mono"`).

### DIFF
--- a/.github/workflows/ci-rust.yml
+++ b/.github/workflows/ci-rust.yml
@@ -63,7 +63,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
-      - run: sudo apt-get update && sudo apt-get install --no-install-recommends -y libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
+      - run: sudo apt-get update && sudo apt-get install --no-install-recommends -y libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev libfontconfig1-dev
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
@@ -83,7 +83,7 @@ jobs:
       - uses: actions/checkout@v5
       - name: Free runner disk space
         run: sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt/ghc /usr/local/.ghcup /opt/hostedtoolcache/CodeQL
-      - run: sudo apt-get update && sudo apt-get install --no-install-recommends -y libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
+      - run: sudo apt-get update && sudo apt-get install --no-install-recommends -y libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev libfontconfig1-dev
       - uses: dtolnay/rust-toolchain@stable
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
@@ -103,7 +103,7 @@ jobs:
       RUSTDOCFLAGS: "-D warnings"
     steps:
       - uses: actions/checkout@v5
-      - run: sudo apt-get update && sudo apt-get install --no-install-recommends -y libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev
+      - run: sudo apt-get update && sudo apt-get install --no-install-recommends -y libasound2-dev libudev-dev libwayland-dev libxkbcommon-dev libfontconfig1-dev
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
         with:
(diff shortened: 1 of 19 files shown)


not-elm:
## Update: independent UI font (`[font].ui`) added

This branch now also adds an **independent UI-chrome font**, layered on top of the family-based `[font]` work above. New commits: `832049a`, `49d255d`, `ed5827a`, `ff80786`, `eb7e436`, `f776a5d`, `a273df8`, `4a5e72f` (plus the small `35657d1` import cleanup).

### What it does
A single `[font].ui` face (`{ family, style }`) sets the font of orzma's own UI (window bar, vi indicator, vi search, IME overlay, tmux rename/confirm prompts) independently of the terminal grid:

```toml
[font]
normal = { family = "Fira Code" }
ui     = { family = "Inter", style = "Medium" }   # UI chrome; omit fields to inherit `normal`
```

- **Per-field inheritance from `normal`**: `ui.family`/`ui.style` each fall back to `normal`'s when omitted (`ui` unset ⇒ UI == terminal `normal`, unchanged behavior).
- **`ui.style` is actually rendered**: `TerminalUiFont` now carries `weight`/`slant`, applied at every UI `TextFont` site via a `text_font(size)` builder.
- **Absent `ui.family` aborts startup** (panic, consistent with terminal faces — no silent fallback).
- **Bundled fallback selects among the 4 bundled faces** by weight/slant, because a single static bundled font can't honor an arbitrary `TextFont.weight`. Intermediate weights (Light/Medium/SemiBold) round to Regular/Bold on that path only; a resolved system family gets the full range via parley.
- Powerline separators, `faces()` (the 4-face terminal set), and terminal-face validation are untouched.

### Quality
- Full workspace suite: **1634 passed, 0 failed** (16 binaries, `--test-threads=1`).
- New unit + integration tests: per-field inheritance, absent-`ui.family` panic, `FontStyleSpec`→bevy_text mapping, bundled 4-face selection, style-only bundled path.
- Built via subagent-driven development: each task spec+quality reviewed, plus a whole-branch review (verdict: ready to merge; no Critical/Important).

PR #290, opened 2026-09-11
feat(orzma_vt): implement HPR, SCOSC/SCORC and the CSI ^ spelling of SD

not-elm:
## Problem

orzma advertises `xterm-256color` when the inherited `TERM` is empty. Step 2 of [`docs/todo/vt-conformance-scope.md`](docs/todo/vt-conformance-scope.md) left three cursor and scroll sequences that `csi_dispatch` still dropped through its catch-all:

| Sequence | Function | Who sends it |
| --- | --- | --- |
| `CSI Ps a` | HPR (character position relative) | `vttest` |
| `CSI s` / `CSI u` | SCOSC / SCORC (save / restore cursor, ANSI.SYS spelling) | blessed's `saveCursorA` / `restoreCursorA`, btop |
| `CSI Ps ^` | SD, in xterm's alternate spelling | xterm's control sequence table; no emitter confirmed |

A program that saved its cursor with `CSI s` and restored it with `CSI u` never came back to where it saved, and a relative column move with `CSI a` did nothing.

## Solution

All three are wiring in `crates/orzma_vt/src/interpreter.rs` onto methods that already exist; no new screen primitive. Nothing outside `orzma_vt` changes.

- **HPR** widens the CUF arm to `(None, b'C' | b'a')`. VT510 stops HPR "at the last position on the line" (p.313) and CUF "at the right border of the page" (p.114). Without DECLRMM both are the last column, and `move_cursor_right` clears the last-column flag the way xterm's `CursorSet` does. When DECLRMM lands, the two stop points have to be split again.
- **SD `^`** gets its own arm onto `scroll_region_down`, without the parameter-count guard `CSI T` needs for XTHIMOUSE. ECMA-48 assigns `05/14` to SIMD, not SD (p.73, p.77); xterm-ctlseqs p.15 keeps the 1991 misprint as an alternate spelling of SD. orzma implements no SIMD and advertises xterm, so it follows xterm, and a `// NOTE:` keeps a future SIMD from taking the arm over. Among the terminals checked, only xterm has this spelling (alacritty, kitty, foot, ghostty and wezterm do not).
- **SCOSC / SCORC** map to `save_checkpoint` / `restore_checkpoint`. VT510 defines both as DECSC / DECRC minus the page number (p.337-338), so they share `ESC 7`'s slot, last-column flag included.
  - Only the **parameterless** spellings act (`params.values().count() == 0`), matching `only_default()` in xterm's `charproc.c`. A DECSLRM-shaped `CSI Pl ; Pr s`, meant for a terminal with left and right margins, must not overwrite the slot `ESC 7` and `?1048` share. kitty and wezterm behave the same; alacritty and foot save regardless of parameters.
  - The kitty keyboard protocol's `CSI > u` / `CSI < u` / `CSI = u` / `CSI ? u` carry a private marker and never reach the SCORC arm.

The scope doc marks the three done and rewrites its "`CSI s` の曖昧性" conclusion to the parameterless rule. It also records one observation for later, not fixed here: xterm reads `CSI 0 T` as XTHIMOUSE, while orzma scrolls.

### Commits

| | |
| --- | --- |
| `b67a006` | HPR |
| `38debab` | SD `^` |
| `09caead` | SCOSC / SCORC |
| `6395fe5` | scope doc |
| `35c7620` | test that a multi-parameter `CSI ^` still scrolls (added in review) |

## Tests

`orzma_vt` goes from 801 to 809 tests, all at the interpreter layer:

- `interpreter/tests/cursor.rs`
  - HPR moves the cursor right by its parameter.
  - HPR stops at the last column.
- `interpreter/tests/line_editing.rs`
  - `CSI ^` and `CSI 0 ^` scroll the region down one row.
  - `CSI 1;2 ^` still scrolls, unlike a multi-parameter `CSI T`.
- `interpreter/tests/cursor_checkpoint.rs`
  - `CSI s` / `CSI u` bracket a detour exactly like `ESC 7` / `ESC 8`.
  - A parameterized `CSI s` does not overwrite the slot `ESC 7` filled.
  - A parameterized `CSI u` restores nothing.
  - Kitty keyboard requests restore nothing.

Five failed before the change (TDD). Three pass before it by design and were checked by mutation: removing the SCORC guard, widening the SCORC arm to `(_, b'u')`, and adding `T`'s guard to the `^` arm each make one of them fail.

`cargo test -p orzma_vt`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all -- --check` pass.

## For the reviewer

- VPR (`CSI Ps e`) is deliberately not here. Whether it stops at the last line or at the bottom margin while DECOM is set is still undecided.
- #291 (DECSET 1034) and #292 (Shift-Tab / Insert) edit other sections of the same scope doc. The three merge cleanly in any order.

### DIFF
--- a/crates/orzma_vt/src/interpreter.rs
+++ b/crates/orzma_vt/src/interpreter.rs
@@ -300,8 +300,8 @@ impl VTActor for Executor<'_> {
                 .device
                 .active_screen_mut()
                 .move_cursor_down(repeat_count(params.value(0))),
-            // CUF
-            (None, b'C') => self
+            // CUF, HPR
+            (None, b'C' | b'a') => self
                 .device
                 .active_screen_mut()
                 .move_cursor_right(repeat_count(params.value(0))),
--- a/crates/orzma_vt/src/interpreter.rs
+++ b/crates/orzma_vt/src/interpreter.rs
@@ -400,6 +400,17 @@ impl VTActor for Executor<'_> {
                     .scroll_region_down(repeat_count(params.value(0)));
                 self.stage(damage);
             }
+            // SD (xterm's alternate spelling)
+            // NOTE: ECMA-48 assigns this final byte to SIMD, not SD. It is
+            // read as SD here because xterm reads it that way, so an
+            // implementation of SIMD must not take this arm over.
+            (None, b'^') => {
+                let damage = self
+                    .device
+                    .active_screen_mut()
+                    .scroll_region_down(repeat_count(params.value(0)));
+                self.stage(damage);
+            }
             // CHT
             (None, b'I') => self
                 .device
--- a/crates/orzma_vt/src/interpreter.rs
+++ b/crates/orzma_vt/src/interpreter.rs
@@ -259,6 +259,18 @@ impl VTActor for Executor<'_> {
                 .device
                 .active_screen_mut()
                 .set_scroll_region(params.value(0), params.value(1)),
+            // SCOSC
+            // NOTE: Only the parameterless spelling saves, as in xterm. A
+            // parameterized `CSI s` is DECSLRM on a terminal with DECLRMM,
+            // and saving on it would overwrite the slot `ESC 7` shares
+            // with `?1048`.
+            (None, b's') if params.values().count() == 0 => {
+                self.device.active_screen_mut().save_checkpoint()
+            }
+            // SCORC
+            (None, b'u') if params.values().count() == 0 => {
+                self.device.active_screen_mut().restore_checkpoint()
+            }
             // DA1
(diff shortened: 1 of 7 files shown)


crates/orzma_vt/src/interpreter.rs
```diff
@@ -400,6 +412,17 @@ impl VTActor for Executor<'_> {
                     .scroll_region_down(repeat_count(params.value(0)));
                 self.stage(damage);
             }
+            // SD (xterm's alternate spelling)
+            // NOTE: ECMA-48 assigns this final byte to SIMD, not SD. It is
+            // read as SD here because xterm reads it that way, so an
+            // implementation of SIMD must not take this arm over.
```
not-elm: Delete its comment.

PR #297, opened 2026-09-12
feat(orzma_vt): implement OSC 4 and OSC 104 so a program can recolor the palette

not-elm:
## Problem

orzma advertises `TERM=xterm-256color`, and that entry advertises `initc`, `ccc` and `oc` — the capabilities a program uses to recolor the terminal's 256-slot palette. None of them worked: `OSC 4` and `OSC 104` fell through `osc_dispatch`'s catch-all, so a theme script's colour changes were silently dropped. `docs/todo/vt-conformance-scope.md` tracked this as a Tier 1 gap — the contract implied by the `TERM` we name.

## Solution

Implements the palette commands end to end in `orzma_vt`; no other crate changes.

- **`OSC 4 ; c ; spec`** sets a slot, and **`OSC 4 ; c ; ?`** reports one, replying `ESC ] 4 ; c ; rgb:rrrr/gggg/bbbb` closed the way the query was — BEL for BEL, otherwise the seven-bit ST.
- **`OSC 104`** resets the slots it names, or the whole table when it carries no number.
- **`RIS`** restores the palette as well, matching xterm's `ReallyReset`.

Decisions worth knowing; the design doc carries the full rationale and the citations.

- The palette stays 8 bits per channel, and Xlib's 16-bit values are narrowed to their high byte. `rgb:` scales by digit count while `#` places digits in the high bits, so `rgb:f` is `0xff` and the red of `#f00` is `0xf0`.
- Only `rgb:` and `#` specs are accepted. Colour names, `rgbi:` and the CIE spaces are refused, because no colour-name database ships with this terminal. **This departs from xterm**, whose manual allows a name.
- Special colours — `OSC 4` at 256-260, and `OSC 5` / `105` / `6` / `106` — are ignored on purpose: no terminfo capability advertises them, and `colorBL` would need SGR blink, which this terminal does not implement. A colour number of 256 or more is dropped rather than truncated to a byte, so it can never overwrite slot 0.
- A malformed pair is dropped on its own and the rest of the command still decodes. **xterm stops at the first error**; Alacritty and kitty continue, and this follows them.

Known constraint: vtparse's 64-parameter cap limits one `OSC 4` to 31 pairs and one `OSC 104` to 63 numbers. Every theme tool checked (pywal, tinted-shell, theme.sh) and terminfo's own `initc` send one pair per command, so nothing in practice reaches it.

Also here: `.claude/rules/rust.md` and the `enumerate-test-cases` skill now reach every manual under `docs/references/` rather than three named files. That is what let this work derive its test cases from `xlib.pdf` and `xterm-ctlseqs.pdf`; the two per-task notes it produced are deleted, per the rule that `docs/todo/` working notes go once the work ships.

### Testing

`cargo test --workspace` — 1982 passed, 0 failed. `cargo clippy --workspace` and `cargo fmt --check` clean. The new coverage runs from the colour-string grammar through the decoder to `Vt::interpret` / `Vt::frame`, and includes a terminator arriving in a later chunk, a set and a query in one command, and the parameter cap.

### Follow-up

`apply_palette_requests` stages a full repaint when a slot changes. Review found those rows are never read downstream — `TerminalGrid::apply` writes only the palette for a rowless frame, and the renderer repacks every cell from the symbolic palette regardless — so raising chunk liveness alone would do. Left out of this PR because it revises a documented contract on the public `Frame` type, and it is worth settling before OSC 10/11/12 inherit the same obligation.


crates/orzma_vt/src/device.rs
```diff
@@ -229,6 +234,46 @@ impl DeviceState {
...
+    ///
+    /// - `OSC 104` with no colour number
+    pub fn reset_indexed_colors(&mut self) -> bool {
+        let indexed = &mut *self.colors.palette.indexed;
+        if *indexed == XTERM_INDEXED {
+            return false;
+        }
+        *indexed = XTERM_INDEXED;
+        true
+    }
```
not-elm, line 263: ロジックはPaletteに記載したい。

crates/orzma_vt/src/device.rs
```diff
@@ -354,9 +399,10 @@ struct Screens {
...
 }
 
-/// The base palette.
-// TODO: Apply the OSC 4 / 10 / 11 / 12 overrides to the carried
-// palette once their handlers land.
+/// The live palette: OSC 4 overrides over the built-in xterm table.
+// TODO: Apply the OSC 10 / 11 / 12 overrides to the carried palette
+// once their handlers land, and restore them in `DeviceState::reset`
+// beside the indexed slots.
 struct ColorTable {
```
not-elm: ColorTableは不要？

crates/orzma_vt/src/device/color.rs
```diff
@@ -130,6 +154,11 @@ impl Palette {
     }
 }
 
+/// The full 256-slot xterm table [`Color::Indexed`] resolves to:
+/// [`ANSI_16`], the 6x6x6 cube on [`CUBE_RAMP`], and the grayscale ramp
+/// from 8 to 238 in steps of 10.
+pub(crate) static XTERM_INDEXED: [Rgb; 256] = build_xterm_indexed();
```
not-elm: Paletteのconstantフィールドに変更してほしい

not-elm: privateにはできない？

not-elm (commented):
FB

PR #304, opened 2026-09-13
feat(orzma_vt): install the wide-character machinery (PR-1 of 4)

not-elm:
## Problem

`orzma_vt` has never known how wide a character is. `Screen::print` advances the cursor by exactly one column per scalar regardless of the glyph, and `Cell` stores a bare `char` with no width and no continuation marker.

The renderer, meanwhile, already measures width itself — `runs_to_cells` walks graphemes and advances by `UnicodeWidthStr::width(...).min(2)`. So the two layers disagree today, and `Run.cols`' own doc ("one column per `char` in `text`") records the VT side of that disagreement. This is a live inconsistency, not a missing feature:

| Input | VT advance | Renderer advance |
|---|---:|---:|
| `あ` | 1 | 2 |
| `é` (NFD) | 2 | 1 |
| `👨‍👩‍👧` | 5 | 2 |

This PR is the first of a four-PR stack. It installs the machinery without switching it on, so that the change that actually makes 「あ」 occupy two columns lands against a codebase that already has somewhere to put it.

## Solution

Nothing outside `#[cfg(test)]` constructs a `CellWidth::Wide` cell yet, so every repair path added below is a runtime no-op. The one behaviour change is the column floor, described at the end. The design is `docs/superpowers/specs/2026-09-13-wide-char-support-design.md` (D1–D10).

**`crates/orzma_vt`**

- `CellWidth::of(char) -> Option<Self>` classifies a scalar as narrow, wide, or zero-width, clamping any reported width above two. `None` is preserved for control characters so the printer can ignore them without re-deriving width.
- `Cell` gains `width: CellWidth` and `extra: Option<Box<CellExtra>>` (a bounded set of at most 9 combining marks) and loses `Copy`. The two `copy_within` calls become `rotate_right` / `rotate_left`, which need neither `Copy` nor `Clone`; the remaining sites take a clone or a reborrow.
- `at_right_edge` splits into `is_last_column()` (position only, which `cursor_parked_past_the_row` uses) and `fits(width)`, the latter written in saturating form so a parked cursor cannot overflow `u16`. `fits` is `#[cfg(test)]` here; the printer starts consulting it in PR-2.
- `Row<Cell>` gains the wide-pair repair primitives — `stamp_at`, `heal_joint`, `normalize_wide_pairs`, `repair_after_resize`, `wide_pairs_intact` — and every cell-writing path routes through them. `resize_cols` repairs a single index rather than sweeping, because it walks history rows on every frame of a window drag. `stamp_at` takes the glyph, its `BodyWidth { Narrow, Wide }`, the pen, and the hyperlink rather than a `Cell`, and mints the cell itself, so a continuation or filler cannot be handed to it at all; a wide body that would run past the row is refused as `StampError::OutOfRow` (a `VtResult`), and `Screen::print` / `DeviceState::print` propagate it as `VtResult<Option<DamageSpan>>`; the interpreter drops a glyph the row refuses.
- `GridSize::new(cols, rows) -> VtResult<GridSize>` is the constructor every caller builds a grid size with (`orzma_vt::error` introduces `VtError` / `VtResult`, with `GridSizeError` as the first variant): a zero axis or one beyond `GridSize::MAX_COLS` / `MAX_ROWS` (4096) is rejected at construction, and the column count is raised to `MIN_COLUMNS = 2`, since a width-2 glyph can never fit a one-column grid.

**`crates/orzma_tty`** — `SpawnOptions` carries a `size: GridSize` instead of `cols`/`rows`, and `resize`, `detached`, and `detached_with_channels` take a `GridSize`, so the one size built with `GridSize::new` reaches both the PTY and the VT and a pane cannot tell its child one width via `TIOCSWINSZ` while the grid holds another. `CellPixels::pty_size(size)` builds the whole `PtySize` from a `GridSize`, so `Pty` never takes separate column and row counts either. The zero-axis / oversized guard that `resize` used to apply (ignore with `Ok`) is gone with it: an invalid size can no longer be built, so the window geometry sender in `src/session/layout.rs` skips the `Resize` instead, and the orzmux backend reports an invalid pane rect as `SpawnFailed` or keeps the old pane size.

**`crates/orzmux`** — the layout's leaf minimum follows the VT floor: `Node::min_size` returns `MIN_COLUMNS` columns for a leaf, and `split()` derives its guard from that minimum (two leaves plus a separator — five columns for a vertical split, three rows for a horizontal one) instead of the hardcoded three. Without this, a one-column pane rect would drive a two-column VT and the grid would be one column wider than its allocation. Two layout tests get new expected geometry from the raised floor; the "below the minimum" test now solves a 1×1 window as a virtual 5×1. `OrzmuxCommand::Resize` carries the window size as a `GridSize` built with `GridSize::new`.

**Workspace** — `unicode-width` pinned to `=0.2.0`. Between 0.2.0 and 0.2.2, 458 codepoints change width (355 of them 1→2), and width is a coordinate shared with every running application.

`orzma_vt` gains 25 tests. `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, and `RUSTDOCFLAGS="-D warnings" cargo doc` all pass.

### Behaviour change

A pane can no longer be one column wide. `MIN_COLUMNS` raises the VT and PTY to two columns, and `orzmux` refuses a vertical split that would leave either child below that (the guard is 5 columns, was 3). Horizontal splits are unaffected.

### Known issues carried into PR-2

Latent today because nothing constructs a wide cell, armed the moment the printer dispatches on width:

- `CellWidth::Spacer` means both "zero-width mark" and "continuation column". The classifier returns it for the former, but the row invariant forbids storing one without a preceding `Wide` — an exhaustive search over intact rows up to length 7 found 246 invariant breaks from stamping one. PR-2 replaces the classifier with a separate `GlyphClass` type.
- A `Wide` glyph stamped into the last column is refused as `StampError::OutOfRow`, so the printer only ever stamps narrow glyphs here. The real fix is the printer consulting `fits`.
- `to_runs` pushes the continuation cell's literal `' '` into the run text while the renderer re-derives advance from grapheme width, so a wide pair would cost three renderer columns for two stored ones.
- `selection_text` reads `.c` from every column, so copying wide text would inject a space after each fullwidth glyph.

### Deliberate deviations from the spec

Three 1-column `Screen` tests are kept as-is rather than widened: they exercise real VT behaviour where column 0 is simultaneously the last column and a deferred wrap arms without the cursor leaving it. Enforcement sits at the `OrzmaVt` boundary, so `Screen::new` is unconstrained and those tests still pass untouched. Consequently no `debug_assert` was added to `Screen::new`/`Grid::new`; the defensive `cols < 2` branch in `print` arrives with PR-2, which is where `print` is restructured.

https://claude.ai/code/session_01RvFA9RWPN6JwMpV1PSoKRS


crates/orzma_tty/src/lib.rs
```diff
@@ -123,14 +123,20 @@ impl<V: Vt> OrzmaTty<V> {
...
             rows: options.rows,
-        });
+        }
+        .normalized();
+        options.cols = size.cols;
+        options.rows = size.rows;
+        let mut tty = Self::wired(vt, Pty::spawn(&options)?);
+        tty.resize_vt(size);
         Ok(tty)
     }
```
not-elm: `SpawnOptions`の`cols`,`rows`の代わりに`GridSize`を渡すようにし、132:137行目の処理を削除してほしい。
また、`GridSize::new`というコンストラクト関数を追加し、その関数内で正規化をしてください。
```rust
impl GridSize {
    pub fn new(cols: u16, rows: u16) -> Self {
        Self {
            rows,
            cols: cols.max(MIN_COLUMNS),
        }
    }
}
```

crates/orzma_tty/src/lib.rs
```diff
@@ -173,15 +181,21 @@ impl<V: Vt> OrzmaTty<V> {
...
-        let pty = Pty::with_master(Box::new(RecordingMaster::at(cols, rows).0), writer);
+        let size = GridSize { cols, rows }.normalized();
+        let pty = Pty::with_master(
+            Box::new(RecordingMaster::at(size.cols, size.rows).0),
+            writer,
+        );
         let mut tty = Self::wired(vt, pty);
-        tty.resize_vt(GridSize { cols, rows });
+        tty.resize_vt(size);
         Ok(tty)
```
not-elm: 以下の指摘と同等
https://github.com/not-elm/orzma/pull/304/changes#r4007635257

crates/orzma_tty/src/lib.rs
```diff
@@ -191,14 +205,15 @@ impl<V: Vt> OrzmaTty<V> {
         chunk_rx: Receiver<Vec<u8>>,
         exit_rx: Receiver<Option<i32>>,
     ) -> OrzmaTtyResult<Self> {
+        let size = GridSize { cols, rows }.normalized();
```
not-elm: 以下の指摘と同等
https://github.com/not-elm/orzma/pull/304/changes#r4007635257

crates/orzma_tty/src/lib.rs
```diff
@@ -265,8 +280,9 @@ impl<V: Vt> OrzmaTty<V> {
         if cols == 0 || rows == 0 || Self::MAX_COLS < cols || Self::MAX_ROWS < rows {
             return Ok(());
         }
-        self.pty.resize(cols, rows, cell_px)?;
-        self.resize_vt(GridSize { cols, rows });
+        let size = GridSize { cols, rows }.normalized();
```
not-elm: 以下の指摘と同等
https://github.com/not-elm/orzma/pull/304/changes#r4007635257

not-elm (commented):
FB

crates/orzma_tty/src/lib.rs
not-elm: Addressed in 647b764c: `GridSize::new(cols, rows)` normalizes and replaces `normalized()`; `SpawnOptions` carries `size: GridSize`, and `resize` / `detached` / `detached_with_channels` take a `GridSize`, so the tty no longer normalizes on its own. The same commit is merged up through #309, #310, and #311.

not-elm: Follow-up in 53e4adcb: the remaining `cols, rows` → `GridSize` sites take the same shape — `OrzmuxCommand::Resize { size: GridSize, cell_px }` (built with `GridSize::new` in the geometry sender), plus the `FakeVt::new` and drain-test frame helpers. Merged up through #309–#311.

crates/orzma_tty/src/lib.rs
```diff
@@ -250,23 +242,23 @@ impl<V: Vt> OrzmaTty<V> {
...
     /// The placements the new geometry strands reach the next pump as a
     /// [`VtSignal::WebviewEvicted`] signal; no PTY output is needed to
     /// carry them.
-    pub fn resize(&mut self, cols: u16, rows: u16, cell_px: CellPixels) -> OrzmaTtyResult {
+    pub fn resize(&mut self, size: GridSize, cell_px: CellPixels) -> OrzmaTtyResult {
+        let GridSize { cols, rows } = size;
         if cols == 0 || rows == 0 || Self::MAX_COLS < cols || Self::MAX_ROWS < rows {
             return Ok(());
         }
         self.pty.resize(cols, rows, cell_px)?;
```
not-elm: このバリデーションですが、`GridSize::new`を`Result`型にして生成の段階でバリデーションを追加するのを検討したい。

not-elm (commented):
FB

crates/orzma_tty/src/pty.rs
```diff
@@ -69,11 +69,13 @@ impl Pty {
...
-        let (pixel_width, pixel_height) = options.cell_px.window_pixels(options.cols, options.rows);
+        let (pixel_width, pixel_height) = options
+            .cell_px
+            .window_pixels(options.size.cols, options.size.rows);
         let pty_pair = native_pty_system()
             .openpty(PtySize {
-                rows: options.rows,
-                cols: options.cols,
+                rows: options.size.rows,
+                cols: options.size.cols,
```
not-elm: `PtySize`の`rows`と`cols`も`GridSize`型のフィールドに置き換えてほしい。

not-elm (commented):
FB

crates/orzma_tty/src/lib.rs
not-elm: Done in 31e15925: `GridSize::new` now returns `Result<Self, GridSizeError>` (thiserror) and rejects a zero axis or one beyond `GridSize::MAX_COLS` / `MAX_ROWS`, which moved from the tty to `GridSize`. This guard in `OrzmaTty::resize` is removed; the window geometry sender skips an invalid `Resize` and the backend reports an invalid pane rect as `SpawnFailed` / keeps the old size. Merged up through #309–#311.

crates/orzma_vt/src/screen/grid/row.rs
```diff
@@ -57,6 +57,156 @@ impl Row<Cell> {
...
+    /// continuation.
+    pub fn stamp_at(&mut self, column: u16, cell: Cell) {
+        let start = usize::from(column);
+        let wide = cell.width == CellWidth::Wide;
+        debug_assert!(
+            matches!(cell.width, CellWidth::Narrow | CellWidth::Wide),
+            "a stamp carries a body width, not a continuation or a filler"
+        );
+        let end = if wide { start + 1 } else { start };
+        debug_assert!(end < self.0.len(), "a stamp stays inside the row");
```
not-elm: 個人的にこれはdebug_assert!でパニックさせるべきではなく適切なエラー型を返すべきだと思います。
返り値の型を`VtResult`にする必要もあります。

crates/orzma_vt/src/screen/grid/row.rs
```diff
@@ -57,6 +57,156 @@ impl Row<Cell> {
...
+    /// `CellWidth::Wide` cell — the row mints its own continuation and
+    /// filler columns — and must leave a `Wide` cell room for its
+    /// continuation.
+    pub fn stamp_at(&mut self, column: u16, cell: Cell) {
+        let start = usize::from(column);
+        let wide = cell.width == CellWidth::Wide;
+        debug_assert!(
+            matches!(cell.width, CellWidth::Narrow | CellWidth::Wide),
+            "a stamp carries a body width, not a continuation or a filler"
+        );
```
not-elm: 型レベルで安全性を担保できないかを調査してほしい。
具体的には,` CellWidth::Narrow | CellWidth::Wide`のような`Run<Cell>`はそもそも`stamp_at`を呼び出せないようにしたい。

not-elm (commented):
FB

crates/orzma_vt/src/screen/grid/row.rs
not-elm: Done in d3dd813e: `Row::stamp_at` returns `VtResult` — a continuation/filler cell is refused as `StampError::NotABody` and a wide body that would run past the row as `StampError::OutOfRow`, leaving the row unchanged — and the assertions are gone. `Screen::print` / `DeviceState::print` propagate it as `VtResult<Option<DamageSpan>>`; the interpreter drops a glyph the row refuses. Merged up through #309–#311.

crates/orzma_vt/src/lib.rs
```diff
@@ -328,10 +331,16 @@ pub struct OrzmaVt {
     tracker: FrameTracker,
 }
 
+const _: () = {
+    const fn assert_send<T: Send>() {}
+    assert_send::<OrzmaVt>();
+};
+
```
not-elm: これの意味がわからない。恐らく不要では？

crates/orzma_vt/src/screen.rs
```diff
@@ -772,13 +782,26 @@ impl Screen {
...
+
+    /// Whether a glyph spanning `width` columns fits from the cursor's
+    /// column through the row's end.
+    #[cfg_attr(
+        not(test),
+        expect(
+            dead_code,
+            reason = "the printer reaches the fit test when width dispatch lands"
+        )
+    )]
```
not-elm: 単純に`#[cfg(test)]`に置き換えられませんか？

not-elm (commented):
FB

crates/orzma_vt/src/screen/grid/row.rs
not-elm: Done in 6284f62d: `Row::stamp_at(column, c, width: BodyWidth, pen: &Pen, hyperlink_id)` no longer takes a `Cell` — it mints the cell itself from `BodyWidth { Narrow, Wide }`, so a continuation or filler cannot be handed to it at all and `StampError::NotABody` is gone (only `OutOfRow` remains). `Pen::stamp` takes a `BodyWidth` too, and the filler has its own `Pen::filler()`. I looked at a private-field newtype, typestate on `Cell`, an enum row storage, and a sealed trait; `BodyWidth` alone gives the same guarantee at this boundary with the least churn, the others would only pay off as a wider storage refactor. Merged up through #309–#311.

crates/orzma_vt/src/lib.rs
not-elm: Removed in 2eb62ed6. It was a leftover from the commit that dropped `Copy` from `Cell`, pinning that `OrzmaVt` stayed `Send`. That is already enforced by the compiler: `OrzmaVt` is moved into the orzmux backend thread inside `OrzmaTty<OrzmaVt>` (`OrzmuxClient::spawn` → `thread::Builder::spawn`), so losing `Send` would fail to build there.

crates/orzma_vt/src/screen.rs
not-elm: Yes — replaced with `#[cfg(test)]` in 2eb62ed6. `fits` has no non-test caller on this PR; PR-2 (#309) drops the attribute when the printer starts consulting it.

PR #305, opened 2026-09-13
feat: implement OSC 52 so a program can write the system clipboard

not-elm:
## Problem

A program running inside orzma had no way to reach the host clipboard. `OSC 52` — the sequence Neovim's osc52 provider, tmux's `set-clipboard`, and SSH-friendly yank helpers use to copy text — fell through `osc_dispatch`'s catch-all, so a yank made over SSH never left the remote host.

Most of the plumbing already existed: `VtSignal::Clipboard` and `TtyClipboardStoreSignal` were defined and forwarded end to end, but `OrzmaVt` never raised the signal and nothing on the GUI side observed it. `docs/todo/vt-conformance-scope.md` tracked this as a Tier 2 gap.

## Solution

Implements the write direction of `OSC 52 ; Pc ; Pd`. The `?` read query is out of scope; see Follow-up.

- **`orzma_vt`**: `ClipboardRequest::parse` in `interpreter/osc/clipboard.rs` decides whether a command sets the clipboard, clears it, or leaves it alone. `osc_dispatch` raises `VtSignal::Clipboard` with the decoded text, or the new `VtSignal::ClearClipboard`. No frame is staged, because nothing on screen changes.
- **Root binary**: a new `ClipboardStoreActionPlugin` (`src/action/clipboard/store.rs`) forwards `TtyClipboardStoreSignal` to the existing `CopyAction`, and `TtyClipboardClearSignal` to a new `ClearClipboardAction`. Both host writes stay in `copy.rs`. `bevy_clipboard` has no clear operation, so a clear writes the empty string, with a `TODO` to switch to a real clear once one is exposed (arboard, which it wraps, has `Clipboard::clear`).
- **`bevy_orzmux`**: adds `TtyClipboardClearSignal` and its arm in `trigger_vt_signal`. `orzma_tty` and `orzmux` are unchanged, since they already forward every `VtSignal` variant.
- **`docs/todo/vt-conformance-scope.md`**: the Tier 2 row and the ordering list record these decisions and what remains.
- Adds `base64 = "0.22"` to `orzma_vt`. The lockfile already resolved 0.22.1 transitively, so nothing new is built.

Decisions worth knowing. They were settled against `xterm-ctlseqs.pdf` p.40–41, Alacritty (`vte` 0.15 / `alacritty_terminal` 0.26), and tmux's `input.c`.

- **Selections.** The host clipboard is the only target, because `bevy_clipboard` has no primary selection. `c`, `s`, and an empty `Pc` — xterm's `s0` default — write it; `p`, `q`, and cut buffers `0`–`7` write nothing. Alacritty on macOS and Windows also drops `p` and `s`. **This departs from Alacritty in two ways**: Alacritty reads only the first byte of `Pc` and defaults an empty `Pc` to `c`, whereas orzma reads `Pc` as the list xterm specifies, so `OSC 52 ; pc ; …` writes and an empty `Pc` means `s0`.
- **A malformed selection list is refused whole.** A `Pc` carrying any byte outside `cpqs01234567` writes nothing. tmux skips such bytes instead, but then `xyz` filters down to an empty list, falls into the `s0` default, and turns garbage into a clipboard write.
- **A bad payload clears the clipboard**, following xterm: "If the second parameter is neither a base64 string nor ?, then the selection is cleared" (`xterm-ctlseqs.pdf` p.40). Invalid or unpadded base64 (`STANDARD`, canonical padding required) and a missing payload field clear it. A payload that is valid base64 but decodes to non-UTF-8 clears it too, since only text can be written. **This departs from Alacritty**, which ignores all of these. An empty `Pd` is valid base64 and sets the clipboard to the empty string.
- **`Pd` is everything after `Pc`.** `OSC 52 ; c ; aGk= ; Zm9v` clears the clipboard, because xterm reads everything after `Pc` as one `Pd` and `;` is not base64. **Alacritty differs**: it takes the third field and ignores the rest.
- A `Pc` naming no clipboard target (`p`, `q`, cut buffers) leaves the clipboard alone whatever `Pd` holds. A bare `OSC 52` with no parameter reads as an empty `Pc` with no `Pd`, and clears the clipboard.
- **A cancelled OSC is ignored, for every OSC.** vtparse closes an OSC on CAN or SUB and dispatches what arrived so far, which would otherwise let a cut-off `OSC 52` write a truncated string or clear the clipboard. `osc_dispatch` now returns early when CAN or SUB closed the command, following ECMA-48 § 8.3.6 ("the data preceding it in the data stream is in error. As a result, this data shall be ignored") and `vt510.pdf` p.63. This also applies to the title, working directory, palette, and dynamic colors. A command interrupted by ESC is still acted on, since that ESC cannot be told apart from the one that opens the string terminator.
- Writes from background panes are accepted, and no setting disables writes. That matches Alacritty's default, `OnlyCopy`.

### Testing

`cargo test --workspace` — 2128 passed, 0 failed. `cargo clippy --workspace --all-targets` and `cargo fmt --check` are clean.

- 14 unit tests in `interpreter/osc/clipboard.rs` pin each rule of `ClipboardRequest::parse`: the selection list, base64 and padding, UTF-8, a missing or empty payload, a bare `OSC 52`, a bad payload for a non-clipboard selection, and `?`.
- 5 interpreter tests in `interpreter/tests/clipboard.rs` cover the reported signal, the ST terminator, a payload that vtparse splits on `;`, a write leaving the chunk undamaged, and a command cancelled by CAN or SUB.
- `interpreter/tests/title.rs` and `interpreter/osc.rs` each gain a test pinning that a cancelled OSC is ignored for commands other than OSC 52, and that `OscTerminator::from_byte` reports CAN and SUB as cancellation.
- 2 tests cover `ClipboardStoreActionPlugin` forwarding the store and clear signals to `CopyAction` and `ClearClipboardAction`.

The `orzma_tty` → `orzmux` → `bevy_orzmux` hop has no new test; those layers forward every signal variant without filtering. This was not exercised in the running app. To try it by hand:

```sh
printf '\033]52;c;%s\007' "$(printf hi | base64)"
```

### Follow-up

- **`?` (read) is not implemented.** Answering from the real host clipboard lets anything that can write to the PTY — including a compromised SSH host — silently read what the user copied, which is why kitty and ghostty prompt before a read, and Alacritty and current xterm keep it off by default. A per-pane buffer that replays only OSC 52 writes would be safe, but it returns stale text to Neovim's `"+p`. Choosing between these, and adding the setting that gates either, is left for a separate change. The config path already reaches `OrzmaVt::new` through `OrzmuxConfig`.
- **Review findings left open**, each needing a decision beyond this change:
  - With no setting to disable writes, a `cat`-ed file can put a command on the clipboard that runs if the user pastes it into a shell without bracketed paste, since unbracketed paste passes control bytes through and turns `\n` into `\r`. Likewise, a malformed OSC 52 in `cat`-ed output now clears the user's clipboard.
  - Every OSC 52 makes its own clipboard write on the main thread, and nothing collapses a burst to the last request, so a program flooding OSC 52 can stall the UI and fill clipboard-history managers.
  - A cancelled APC is still acted on. vtparse closes APC strings on CAN and SUB too and calls `apc_dispatch` with the partial data, so an `ESC _ Ounmount;n=<id>` cut off right after `Ounmount` parses as an unmount of every webview. This predates this change and alters behavior, so it is left out.
  - There is no payload size cap. vtparse buffers OSC data without limit under `std`, which is also parser-wide.


crates/orzma_vt/src/interpreter/tests/clipboard.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Asserts that a payload carrying a `;` is refused whole as not
+/// base64, rather than read up to the first `;`.
+///
+/// Case: a script joins two encoded strings with `;`, so the payload
+/// runs on past the first one.
+#[test]
+fn a_payload_carrying_a_semicolon_reports_nothing() {
+    let (_device, output) = interpret_fully(b"\x1b]52;c;aGk=;Zm9v\x07");
+    assert!(output.signals.is_empty());
+}
```
not-elm: XTermの仕様上Clearするのが正しいため`VtSignal::ClearClipboard`を追加し、それをpushするようにして。
`bevy_clipboard`にclearが存在しない可能性があるため当分空文字で代用

PR #307, opened 2026-09-13
feat(orzma_vt): implement OSC 8 hyperlinks

not-elm:
## Problem

`OSC 8` hyperlinks reached `osc_dispatch` and fell through to `_ => {}`. A program could emit `OSC 8 ; params ; URI ST` and nothing happened: no cell carried a link, no frame carried a target, and nothing was clickable. `docs/todo/vt-conformance-scope.md` listed it under Tier 2 as the sequence nvim relies on.

Everything downstream of the VT was already built and had been waiting on it. `TerminalGrid` kept a retained id-to-target table and resolved a link per rendered run, `src/input/hyperlink.rs` tracked Cmd-hover, and `src/action/terminal/open_uri.rs` opened an allowlisted target on Cmd-click. `Run::hyperlink_id` was documented "it is always `None`" and `Frame::hyperlinks` "It is always empty" — the two sentences this PR makes false.

## Solution

The VT side of the path, in the order the data moves:

```
OSC 8 bytes
  -> HyperlinkRequest::parse      interpreter/osc.rs, a pure decoder
  -> DeviceState::open_hyperlink  device.rs, interns and writes the pen
  -> HyperlinkInterner            hyperlink.rs, (source id, target) -> HyperlinkId
  -> Pen.hyperlink_id             cell.rs, per screen
  -> Cell.hyperlink_id            cell.rs, burned by Pen::stamp
  -> Run.hyperlink_id             screen/grid/row.rs, to_runs splits on it
  -> Frame.hyperlinks             frame.rs, resolves the ids the repainted rows reference
```

`HyperlinkId` now wraps a `NonZeroU32`. Zero was documented as the "no hyperlink" sentinel but nothing stopped a caller building it, and the GPU path already encodes zero the same way; making it unrepresentable also fits `Option<HyperlinkId>` into four bytes, so a `Cell` grows from 16 to 20 rather than 24. A `const` assertion pins that.

A target arrives split on every `;` because a URI may legally contain one, so the decoder rejoins the tail. An `id=` is a lookup key, and an absent or empty one takes a fresh id so two such links never join — that rule lives in `HyperlinkInterner::open` alone, and the decoder passes the parameter through as written.

Ids are monotonic and never reused. `RIS` clears the pen on both screens but deliberately keeps the interner: the renderer only ever adds unknown ids to its table and never replaces a known one, so a reused id would resolve to the target it carried before the reset. That hazard is recorded where the omission would otherwise look like a forgotten line.

No `SGR` clears a link, `SGR 0` included — a program that ends a coloured link with `ESC[0m` before closing it would otherwise lose the rest. `Pen::applied` restores the field around the whole SGR walk rather than inside the reset arm, so a future arm that replaces the whole pen cannot silently drop it.

**Affected:** `crates/orzma_vt` (the feature), `crates/bevy_orzma_tty_renderer` (the two consequences below), `src/input/hyperlink.rs` (test call sites), `docs/todo`.

### Two paths this PR activates rather than introduces

Both were unreachable while `Run::hyperlink_id` was always `None`, so wiring `OSC 8` is what makes them run. Neither is visible in a diff of the VT work alone.

The shader painted an underline for any linked cell and hover-gated only the accent colour. That is kept — links are underlined whenever they are on screen — but the two branches that stated the same band geometry are folded into one. Gating the paint on hover was tried and reverted: it leaves a link with nothing to mark it until the pointer happens to cross it with the modifier held.

`runs_to_cells` cloned the whole target into every grapheme cell of a linked run on every repaint, and `FrameTracker::definitions` cloned one per referenced link per frame, while the renderer's retained table was a `Vec` scanned linearly once per run per repaint and never pruned. `HyperlinkUri` now wraps an `Arc<str>` so both clones are refcount bumps, and the table is a `HashMap` keyed by id. The public surface, the derives and `Send + Sync` are unchanged.

### Behaviour change beyond the new sequence

A flip onto the alternate screen now closes whatever hyperlink that screen's pen still held. The pen carries over between uses on purpose, which is right for colour but not for a clickable target: a full-screen program killed before it emitted the closing sequence left every cell the next one printed opening the dead program's target on a click, and the frame re-emitted the definition so the renderer resolved it. Both `1049` and `1047` are covered; the rest of that pen is untouched.

### Not in this PR

No length or byte-range validation of the target, and no reclamation — the interner grows with the `OSC 8` sequences a program sends rather than the links it puts on screen, and nothing releases an entry, not even a full reset. Reclamation needs a design of its own (when to release, how that meets scrollback cells still referencing an id, and how the renderer's table learns of a removal), so it is written up as item 8-b in `docs/todo/vt-conformance-scope.md` rather than guessed at here.

Also out of scope: reference counting, the iTerm2 reading that joins adjacent id-less links sharing a target, `file://` host validation (`is_allowed` does not list the scheme, so those targets never open), and wide characters, which are unimplemented crate-wide.

### Testing

`cargo test --workspace --no-fail-fast` is 2104 passed / 2 failed, the two being the environment-dependent `font` tests that also fail on a clean checkout. `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` are clean. The WGSL change was validated with the `naga` CLI through parse, typecheck and MSL codegen — that is syntax and types, not rendered pixels.

Around thirty tests were added. The decoder is covered at the unit level, including the cases only reachable there. The interactions with `DECSC`/`DECRC`, `DECSTR`, `RIS`, erase, the alternate screen and automatic wrap are pinned as characterization tests, as is `vtparse`'s 64-parameter cap truncating a target carrying many semicolons.

Three tests were found to assert something the code could not violate and were rewritten until reverting the relevant change made them fail. That check is worth repeating on any test here that looks like it only ever passes.

<details><summary>Not verified</summary>

The live Cmd-hover and Cmd-click path was never exercised — no GUI automation was available. `hyperlink_id` and `hover_active` are uniform fields that go live for the first time in this PR, so a wrong offset or a stale pipeline cache would show as a missing or misplaced underline rather than a test failure.

```
printf '\e]8;;https://example.com\e\\link\e]8;;\e\\\n'
```

`ls --hyperlink=auto` is not usable for the click check: coreutils emits `file://<host>/<path>`, which `is_allowed` rejects, so those links underline and change the pointer but never open.

</details>

https://claude.ai/code/session_01WDs2FsrK2AserhJEra866h


crates/orzma_vt/src/screen/grid/row.rs
```diff
@@ -34,13 +34,19 @@ impl Row<Cell> {
...
     pub fn to_runs(&self) -> Row<Run> {
         let mut runs: Vec<Run> = Vec::with_capacity(self.0.len().min(Self::RUNS_RESERVE));
         for cell in self.0.iter() {
             match runs.last_mut() {
-                Some(run) if run.fg == cell.fg && run.bg == cell.bg && run.style == cell.style => {
+                Some(run)
+                    if run.fg == cell.fg
+                        && run.bg == cell.bg
+                        && run.style == cell.style
+                        && run.hyperlink_id == cell.hyperlink_id =>
```
not-elm: Runのconstメソッドとして抽出したい。

crates/orzma_vt/src/screen/cell.rs
```diff
@@ -17,15 +18,23 @@ pub struct Cell {
...
     pub style: Style,
+    /// The hyperlink the glyph was printed inside, if any.
+    pub hyperlink_id: Option<HyperlinkId>,
 }
 
+const _: () = assert!(
+    size_of::<Cell>() <= 20,
+    "a cell must not grow past twenty bytes"
+);
+
```
not-elm: 削除して

crates/orzma_vt/src/screen/cell.rs
```diff
@@ -62,13 +71,15 @@ impl Default for Pen {
 }
 
 impl Pen {
-    /// Burns the pen's attributes into a cell holding `c`.
-    pub fn stamp(&self, c: char) -> Cell {
+    /// Burns the pen's attributes into a cell holding `c`, printed inside
+    /// `hyperlink`, or outside any link when it is `None`.
+    pub fn stamp(&self, c: char, hyperlink: Option<HyperlinkId>) -> Cell {
```
not-elm: 引数の名前を`hyperlink_id`にして

not-elm (commented):
FB

PR #308, opened 2026-09-14
feat(input): resize panes by dragging the divider between them

not-elm:
## Problem

The divider between two panes was decorative. `orzmux` solved the split tree with a fixed `ratio` of 0.5 that only `LayoutTree::split` ever wrote, so the only way to change how much room a pane got was to close it and split again. Nothing in the GUI could address a divider either: `Separator` carried `orientation`, `x`, `y`, `len` and no identity, and `bevy_orzmux` matched separator entities to layout entries by list index, so "which divider is this" had no answer that survived a tree mutation.

## Solution

Dragging a divider resizes the two panes it separates, live, at cell granularity. All geometry stays in the Bevy-free backend; the GUI hit-tests a grab band and asks for a cell.

- **`orzmux`**: `LayoutTree` mints a `SplitId` per split node and `solve_into` carries it on every `Separator`. `Split::subdivide` and `LayoutTree::root_rect` factor out the per-split geometry the solver already computed, and the new `resize_split` reuses both rather than re-deriving them — if the two ever disagreed, the GUI's coordinate space and the backend's would silently diverge. `Node::min_size_for_drag` is the drag-only companion to `min_size`; `min_size` itself is unchanged, so window-shrink behaviour is untouched. `OrzmuxCommand::ResizeSplit { split, position }` carries a `u16` cell and nothing else, so the channel types still cross a socket unchanged.
- **`bevy_orzmux`**: `OrzmuxSeparator` becomes a public component carrying `split` and `orientation`, and `reconcile_separators` is re-keyed from list index to `SplitId`. `CurrentLayout` deliberately stays `pub(crate)` — the host reads identity off the entity, so it never needs the layout resource. `RequestSplitResize` is the host-facing event, gated like every sibling on `resource_exists::<OrzmuxConnection>`.
- **`src/input`**: a new `src/input/mouse/separator.rs` owns `SeparatorHit::resolve` (the grab band), `GrabbedSeparator` (the held divider), `drive_separator_drag` and `retire_separator_drag`. A `MousePhase { Grab, Dispatch, Retire }` set, nested inside `InputPhase::Dispatch`, orders the grab ahead of `dispatch_mouse_buttons` and `route_webview_pointer`, both of which now stand down while a divider is held. `hyperlink.rs` gains a `HoverTarget::Separator` arm so the pointer shows `ColResize` / `RowResize`.

Decisions worth knowing.

- **The backend owns every clamp.** The GUI sends an absolute whole-window cell, never a delta, and `resize_split` converts it to a ratio and clamps it. That keeps `f32` off the wire, and it means a drag that overshoots and comes back needs no hysteresis handling — the cursor position is always the truth.
- **`position` is the divider's cell, not the boundary under the cursor.** `separator_node` paints the line at the *far* edge of the reserved cell, so cell `k` puts the line on the `k`/`k+1` boundary. `boundary_at` therefore rounds to the nearest boundary and subtracts one. Sending "the cell the cursor is in" instead makes the line jump a cell the moment you grab the right half of the band.
- **No optimistic apply.** The divider follows the confirmed `Layout`, one or two frames behind the pointer. Applying early would mean duplicating the clamp in the GUI, and the pane contents are backend-rendered anyway, so only the line would move.
- **The grab is a component on the separator entity**, not a resource. It vanishes with the entity when a shell exits and the split collapses, so there is no teardown path for that case. It carries `split` and `orientation` copied at grab time; `SplitId` is never reused and `reconcile_separators` keys by it, so the copy cannot drift, while re-querying every frame would leave a path where an `Err` strands the grab with the divider stopped.
- **The suppression depends on a sync point.** `GrabbedSeparator` is inserted through `Commands` during `MousePhase::Grab`, and Bevy's auto-inserted `ApplyDeferred` on the `Grab → Dispatch` edge is what makes it visible to the dispatchers *in the same frame*. `.ignore_deferred()` on that chain would remove the sync point with no compile error and no panic — just a selection starting in the neighbouring pane. There is a `// NOTE:` saying so, and a test that turns red if the production chain is mutated that way.
- **A same-frame press and release is the ordinary case**, not an edge case: it is one click on a divider. The insert seeds `releasing` from the press frame's observed release, and `retire_separator_drag` tears down one phase later. Without both, a single click would leave the grab alive forever and suppress every later click in the window.
- **The grab band is measured in physical px** and is `max(4 logical px × scale, half a cell)` either side of the painted line, bounded on both axes by the divider's own length.

### Known limitations

Recorded in the spec's scope-out section rather than fixed here:

- **Dividers are not draggable while an IME composition is open**, or whenever every pane is `MouseDisabled`. The cursor still shows the resize shape, so the two disagree. This is the most likely one to be noticed in daily use.
- **The band takes about half a cell from the pane after the divider.** A press there starts a (zero-length) drag instead of reaching that pane, so it sets no selection anchor, does not focus the pane, and sends no mouse report to a program using `set mouse=a`.
- **Pointer motion still reaches an inline webview during a drag.** `route_webview_pointer` and `forward_webview_mouse_moves` both stand down now, but the latter is registered in `InputPhase::Hover`, which runs *before* `MousePhase::Grab` — so on the press frame itself the grab does not exist yet and that frame's `CursorMoved` is forwarded. The follow-up below closes this.
- **The wheel is not gated.** A wheel notch during a drag still scrolls whatever pane is under the pointer. Gating it is a user-visible behaviour change with no bug attached to it.
- **The resize cursor can persist after release** if the pointer ends away from every band and no further input arrives. `InputPhase::Hover` runs before `MousePhase::Retire`, so the release frame still sees the grab; the next frame carries no message at all. Adding a button-message clause to the run condition does not help — it was tried and reverted.
- **A `[Release, Press]` pair coalesced into one frame loses the press.** `left_button_edges` reduces a batch to two order-blind booleans. An order-aware rewrite would have to insert a second grab while the first is still queued for removal.
- **Live resize does not coalesce across commands.** Dragging a divider 35 cells sends 35 `ResizeSplit` commands, each resizing the PTYs whose size changed. A TUI in either pane reflows once per cell crossed.

### Follow-up

Registering `drive_separator_drag` `.before(InputPhase::Hover).after(maintain_input_gates)` — about five lines — would make all three stand-down guards accurate on the press frame at once, closing the webview leak above. It needs one test that drives a press and a `CursorMoved` in the *same* `app.update()`; none of the current tests do, which is why the hole is invisible today. A `PointerCapture` resource was considered and rejected: the hole is an ordering fact, not a storage fact, so a resource read in `InputPhase::Hover` would see `None` on the press frame just the same.

### Testing

`cargo test --workspace` — 342 passed in the `orzma` binary, plus every other target green. The two `font::tests` failures on this machine are pre-existing and environment-dependent; they fail identically on `main`. `cargo clippy --workspace --all-targets -- -D warnings`, `RUSTFLAGS="-D warnings" cargo build --workspace` and `cargo fmt --check` are all clean.

- **`orzmux`** (53 tests): `resize_split`'s cell → ratio → cell round trip, the drag clamp on both axes, the fallback to the tree minimum when a window cannot honour the drag minimum on both sides, a saturating position before the split's origin, a nested split leaving its ancestor alone, `SplitId` surviving an unrelated subtree collapse and never being reused, and the backend publishing exactly one layout per accepted resize and nothing at all for a refused one.
- **`bevy_orzmux`** (37 tests): separator entities following their `SplitId` rather than their position in the list — asserted on the `Node` each entity carries, because the pairing alone holds even for the index-keyed implementation this replaces.
- **`src/input`**: the grab band on both orientations including the along-axis bound and the T-junction tie-break, the cursor → cell-boundary conversion, a drag requesting one move per boundary and nothing while it stays put, the off-window continuation, and the suppression of both dispatchers — including a press and release arriving in one frame, and a positive control asserting that a press *away* from every divider still reaches the dispatcher, so the suppressed cases are suppression rather than an inert harness.

Fixtures run at scale 2.0 wherever a conversion would otherwise be the identity. At scale 1.0 with an 8 px cell the band's two terms are both 4.0, which makes the `× scale` factor unobservable — an earlier revision of these tests passed with the band constant set to `0.0`.

**Not exercised in a running window.** Every test uses `MinimalPlugins` with a synthetic `Window` and hand-built `ComputedNode` / `UiGlobalTransform`, so three things agree with the tests by construction rather than by observation: that a separator node's transform really is the painted line's centre, that the pane container's origin coincides with whole-window cell (0, 0), and that the whole round trip closes with the line under the pointer. To check by hand:

```
split the window, press on the divider, and drag it slowly across several cells;
then split one pane the other way and drag the horizontal divider too.
```

If the line tracks the pointer within a cell in both directions rather than drifting or running at half speed, all three are settled. Dragging past the window edge also answers whether winit keeps delivering `CursorMoved` while the button is held outside the client area — if it does not, the divider freezes at its last in-window position rather than misbehaving.

https://claude.ai/code/session_011GGKBhB3ARfCiCQ4hu5gbo


crates/bevy_orzmux/src/layout.rs
```diff
@@ -107,32 +128,43 @@ fn apply_layout(
 /// parenting new ones under `container` when the host has marked one.
 fn reconcile_separators(
     commands: &mut Commands,
-    separators: &mut Query<(Entity, &mut Node), (With<OrzmuxSeparator>, Without<OrzmuxPane>)>,
+    separators: &mut Query<
+        (Entity, &mut Node, &OrzmuxSeparator),
+        (With<OrzmuxSeparator>, Without<OrzmuxPane>),
```
not-elm: ` (With<OrzmuxSeparator>, Without<OrzmuxPane>)`は不要のはず。

crates/bevy_orzmux/src/layout.rs
not-elm: `container`が`None`になるケースは基本的に存在しないはず？のため`Option`を外したい。
`unwrap`はなく以下のようなパターンマッチが望ましい。
```rust
let Ok(container) = container.single() {
return;
}
```

crates/bevy_orzmux/src/layout.rs
```diff
@@ -84,7 +102,10 @@ fn apply_layout(
     mut commands: Commands,
     mut registry: ResMut<PaneRegistry>,
     mut nodes: Query<&mut Node, With<OrzmuxPane>>,
-    mut separators: Query<(Entity, &mut Node), (With<OrzmuxSeparator>, Without<OrzmuxPane>)>,
+    mut separators: Query<
+        (Entity, &mut Node, &OrzmuxSeparator),
+        (With<OrzmuxSeparator>, Without<OrzmuxPane>),
```
not-elm: (With<OrzmuxSeparator>, Without<OrzmuxPane>)は不要のはず。

not-elm (commented):
FB

src/input/mouse/webview/router.rs
```diff
@@ -60,9 +62,14 @@ fn route_webview_pointer(
         ),
         With<OrzmaTerminal>,
     >,
+    grabbed: Query<(), With<GrabbedSeparator>>,
     metrics: Res<TerminalCellMetricsResource>,
     windows: Query<&Window, With<PrimaryWindow>>,
 ) {
+    if !grabbed.is_empty() {
```
not-elm: System内で`grabbed.is_empty()`をチェックするのではなく以下のようにRun Conditionを作成する方針にできませんか？その場合`buttons.clear()`がなくなりますが、恐らく不要だと思うのですが調査してほしい。
`grabbed.is_empty() `を使用しているすべてのSystemが対象

```rust
app.add_systems(Update, route_webview_pointer
.run_if(not(is_dragging_any_separator))
.in_set(MousePhase::Dispatch))
...

fn is_dragging_any_separator(grabbed: Query<(), With<GrabbedSeparator>>) -> bool{
 !grabbed.is_empty()
}
```

not-elm (commented):
FB

src/input/hyperlink.rs
```diff
@@ -168,17 +174,82 @@ fn write_cursor_icon(
...
+    fn unlocated(held: Option<SplitOrientation>) -> Self {
+        Self::resolve(held, None, || Self::Default)
+    }
+
+    /// The region for the topmost mouse-enabled surface under
+    /// `cursor_phys`, in window physical px. Records that surface and
+    /// the hyperlink id of the cell under the pointer in `hover`,
+    /// leaving both untouched over anything but a terminal grid.
+    /// `cell_phys` is the `(width, height)` cell pitch in physical px.
+    fn over_surface(
```
not-elm: あまり純粋なメソッドに`Query`などのBevyのSystemStateを渡したくない。
このような複数のSystemStateが依存する処理は専用のSystemParam構造体を定義したい。
`resolve`,`unlocated`についてもSystemParamに移動した方がいいかも検討したい。

src/input/hyperlink.rs
```diff
@@ -49,80 +51,84 @@ pub(crate) fn link_modifier_held(mods: &Modifiers) -> bool {
...
+>;
+
 /// Skips any surface with input suppressed (`MouseDisabled`), so hover
-/// never advertises a link the mouse dispatcher would refuse to open.
+/// never advertises a link the mouse dispatcher would refuse to open. A
+/// divider the pointer holds or hovers claims the cursor before any
+/// surface is read, leaving the hover state empty; a held divider keeps
+/// the cursor even while the pointer reports no position, which is what
+/// a drag past the window's edge does.
 fn hyperlink_hover_and_cursor(
```
not-elm, line 74: このSystemの引数が多すぎるため分離できないかを検討したい。

not-elm (commented):
FB

PR #309, opened 2026-09-14
feat(orzma_vt, bevy_orzma_tty_renderer): display fullwidth glyphs across two columns (PR-2 of 4)

not-elm:
## Problem

After #304 the VT knows how wide a character is but still advances one column per scalar, and the renderer still measures grapheme width on its own. The two layers disagree on every fullwidth glyph and every combining mark: `あ` takes one VT column and two renderer columns, `é` (NFD) two VT columns and one renderer column. Japanese text renders shifted, the block cursor covers half a kanji, and copying `あい` yields `あ い`.

## Solution

This PR switches the machinery on and makes the VT the single width authority.

**`crates/orzma_vt`**

- `GlyphClass { Narrow, Wide, ZeroWidth }` replaces `CellWidth::of` as the classifier — a separate type from the stored `CellWidth`, so a zero-width mark can never be written to a cell as a width. It is the workspace's only `unicode-width` call site and is exported through the prelude for the renderer and the IME overlay. `MAX_COMBINING` becomes a public top-level constant.
- `Screen::print` dispatches on width: a fullwidth glyph stamps a `Wide` body and its `Spacer` continuation, a zero-width mark attaches to the cell the cursor last passed (`attach_zero_width`), a wide glyph that does not fit wraps first and leaves a `LeadingSpacer` filler — or, with DECAWM reset, is dropped with the deferred wrap explicitly disarmed. IRM shifts by the glyph's width. Cursor advance has a single exit; a wrap reports both rows as damaged.
- `Run` gains `widths: Vec<u8>` — one entry per `char` of `text` (`0` = mark, `1`, `2`), lazily created so an ASCII row still allocates nothing. `Spacer` cells emit nothing; `LeadingSpacer` emits one blank.
- `Run::check` validates that contract and returns `VtError::Run(RunError)` (`WidthCount` / `WidthSum` / `InvalidWidth` / `LeadingContinuation`) instead of relying on `debug_assert!`s in the consumer.
- `selection_text` skips continuation columns and fillers (the mark-preserving copy lands in PR-3).
- A `proptest` property test (512 generated print/edit/resize sequences) asserts every row, history included, keeps the wide-pair invariant after every op.

**`crates/bevy_orzma_tty_renderer`**

- `runs_to_cells` walks `text.chars().zip(widths)` instead of re-measuring graphemes; width-0 chars join the previous `GridCell.text`, which is the input contract PR-4 renders. `unicode-segmentation` and `unicode-width` leave the crate.
- `TerminalGrid::apply` returns `VtResult`: it runs `Run::check` over every run before touching the grid, so a malformed frame is rejected with the grid left untouched, and the frame observer logs the rejection.
- The cursor shader's `cursor_covers` makes the block and underline cursors cover both halves of a wide glyph in both directions (cursor on the body, or parked on the right half by CHA/CUF); the bar cursor stays exact-cell.

**`src/ui/ime_overlay`** — the preedit measures per `char` with `GlyphClass`, so it agrees with the grid (`👍🏻` is 4 cells in both). `unicode-segmentation` leaves the workspace entirely.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, and `RUSTDOCFLAGS="-D warnings" cargo doc` all pass.

### Behaviour change

Width is now computed per Unicode scalar, matching `wcwidth`-based applications (shells, vim, tmux). Emoji sequences change width as a deliberate consequence: `❤️` (U+2764 U+FE0F) takes 1 column, `👍🏻` 4, `👨‍👩‍👧` 6. Grapheme-cluster width (mode 2027 / OSC 66) is out of scope for this stack.

### Known limitation carried into PR-4

Combining marks reach the renderer in `GridCell.text` but are not yet drawn: `e` + U+0301 occupies one cell and copies correctly, but the screen shows `e`.

https://claude.ai/code/session_01FconFncbHZjBeiaKStVf6X


crates/bevy_orzma_tty_renderer/src/schema/grid.rs
```diff
@@ -287,16 +286,36 @@ fn runs_to_cells(
...
+        );
+        debug_assert!(
+            run.widths.is_empty()
+                || run.widths.iter().map(|w| u32::from(*w)).sum::<u32>() == u32::from(run.cols),
+            "a run's widths sum to its columns"
+        );
+        debug_assert!(
+            run.widths.iter().all(|w| *w <= 2) && run.widths.first() != Some(&0),
+            "a run's widths are 0, 1 or 2 and never start with a continuation"
+        );
```
not-elm: これらは全て`Err`として返すべき処理です。

not-elm (commented):
FB

crates/bevy_orzma_tty_renderer/src/schema/grid.rs
not-elm: 対応しました (66c0b9dd)。契約の検査は `Run::check(&self) -> VtResult` として orzma_vt 側に置き、`VtError::Run(RunError)` (`WidthCount` / `WidthSum` / `InvalidWidth` / `LeadingContinuation`) を返します。`TerminalGrid::apply` はグリッドに触れる前に全 run を検査して `VtResult` を返すので、不正なフレームはグリッドをそのままにして拒否され、observer が warn を出します。`runs_to_cells` の `debug_assert!` は削除しました。

crates/orzma_vt/src/screen/grid/row.rs
```diff
@@ -38,23 +38,36 @@ impl Row<Cell> {
...
                     style: cell.style,
-                    text: cell.c.to_string(),
+                    text: String::new(),
+                    widths: Vec::new(),
                     hyperlink_id: cell.hyperlink_id,
-                }),
+                });
+                chars_in_run = 0;
+            }
+            let run = runs.last_mut().expect("the row has a run to extend");
```
not-elm: ここを安全な処理にしたい。

crates/orzma_vt/src/screen/grid/row.rs
```diff
@@ -98,6 +111,22 @@ impl Row<Cell> {
         Ok(())
     }
 
+    /// Stamps the last column as the blank a wide glyph leaves behind
+    /// when it does not fit there, carrying `pen`'s attributes, and
+    /// restores the wide-pair invariant to its left.
+    pub fn place_filler(&mut self, pen: &Pen) {
+        debug_assert!(!self.0.is_empty(), "a filler needs a column");
```
not-elm: `Err`を返すようにして。
debug_assert!はなるべく利用せずにリザルト型でエラーハンドリングする旨をルールファイルに追記して。

crates/orzma_vt/src/screen/grid/row.rs
```diff
@@ -98,6 +111,22 @@ impl Row<Cell> {
...
+        let last = self.0.len() - 1;
+        self.0[last] = pen.filler();
+        if last > 0 {
+            self.heal_joint(last - 1);
+        }
+        debug_assert!(
+            last == 0 || self.joint_intact(last - 1),
+            "a filler left a broken joint"
+        );
```
not-elm: Errを返して

not-elm:
debug_assert!やexpect. unwrapはなるべく利用せずにリザルト型でエラーハンドリングする旨をルールファイルに追記して。

not-elm:
対応しました (9f61d460)。`.claude/rules/rust.md` に「Error handling — return `Result`, don't assert or unwrap」を追加しました。前提条件や欠損値は `debug_assert!` / `unwrap` / `expect` / `panic!` ではなく `thiserror` の enum を伴う `Result` で返し、回復できない境界 (Bevy の system / observer など) で warn して状態を保つ、型で表せる前提条件は型にする、という内容です。例外はテストコードと、直前に検査済みで成立が自明な条件 (`// NOTE:` + `#[expect(clippy::expect_used)]`) のみ。既存コードは触ったときに順次変換する扱いにし、Enforcement の review-time リストと CLAUDE.md の参照にも追記しました。

crates/orzma_vt/src/screen/grid/row.rs
not-elm: 対応しました (cd54142a)。延長中の run を `Option<Run>` に持ち、属性が変わるところで `runs` に押し出して `get_or_insert_with(|| Run::opened_by(cell))` で次の run を開く形にしたので、`runs.last_mut().expect(..)` は無くなりました。run の切れ目・widths・marks の扱いは変わらず、既存の `to_runs` テストがそのまま通っています。

crates/orzma_vt/src/screen/grid/row.rs
not-elm: 対応しました (72de3a72)。`place_filler` は `VtResult` を返し、列の無い行は `StampError::OutOfRow`、heal 後もジョイントが壊れていれば新設の `StampError::BrokenJoint` で返します。`Screen::print` は `?` で伝播します。ルールファイルへの追記は 9f61d460 で済んでいます。

crates/orzma_vt/src/screen/grid/row.rs
not-elm: 対応しました (72de3a72)。`place_filler` は `VtResult` を返し、列の無い行は `StampError::OutOfRow`、heal 後もジョイントが壊れていれば新設の `StampError::BrokenJoint` で返します。`Screen::print` は `?` で伝播します。ルールファイルへの追記は 9f61d460 で済んでいます。

PR #314, opened 2026-09-15
feat(input): forward the mouse wheel to mouse-tracking apps and alternate scroll (PR-2 of 2)

not-elm:
## Problem

The mouse wheel does not scroll in nvim, `less`, or `man`. The wheel dispatcher always scrolled the local scrollback. Those programs run on the alternate screen, which has no scrollback, so the wheel did nothing there. The app-forwarding path was removed in the engine swap (#268) and never came back (`TODO: reintroduce app-forward wheel-reporting`).

This is PR-2 of 2, stacked on PR-1 (#313), which reports each pane's VT modes to the GUI as a `TtyModes` component.

## Solution

The dispatcher now reads the pane's `TtyModes` and routes each vertical notch in priority order:

1. **Mouse tracking on, Shift not held** (nvim's `mouse=nvi` sends DECSET 1002/1006): one wheel report per notch (`cb` 64/65), sent through `RequestTtyMouseInput`. The app decides how far a notch scrolls.
2. **Alternate scroll in effect** (DECSET 1007 on the alternate screen, e.g. `less`, `man`): the notches' lines become cursor-up/down keys, sent through `RequestTtyKeyInput`. The keys honour DECCKM, as xterm's alternate scroll does.
3. **Otherwise:** the viewport scrolls, as before.

Horizontal notches only produce wheel-left/right reports (`cb` 66/67) while tracking is on.

Details:

- **Shift falls through one level.** Over nvim it sends cursor keys; over a primary-screen tracking app such as fzf it scrolls the scrollback. It is never a dead key.
- **macOS Shift+wheel.** macOS turns Shift+wheel into horizontal travel for discrete wheels only; winit reports those as line units. On macOS the dispatcher adds a Shift-held line-unit event's horizontal delta back onto the vertical axis and clears the horizontal sub-notch residual. Trackpad (pixel-unit) travel keeps its axes. On macOS, `fine_modifier = "shift"` therefore no longer selects fine scrolling.
- **Horizontal sign.** `MouseWheel.x` is negated on every platform. winit reports a positive x as content moving right, which scrolls toward the left, on every backend.
- **Burst cap.** `max_protocol_events_per_frame` (default 8) caps the notches per routing call on both the report route and the cursor-key route. Excess notches are dropped rather than carried into later frames. The cursor-key route is also capped at 240 keys per call, whatever `lines_per_notch` is set to.
- **Report modifiers.** Shift, Ctrl and Alt map to the report's modifier bits; a held Super (Cmd) sets none.
- **`orzma_vt`.**
  - Alternate scroll becomes an `AlternateScroll` enum that defaults to **enabled**. RIS restores the default.
  - `VtModes::mouse_reporting_active()` is added.
  - The private-mode dispatch names the mouse modes it forwards (`1000 | 1002 | 1003 | 1005 | 1006`) instead of sending every unknown number to the mouse handler. Behaviour is unchanged.
- **`orzma_tty`.**
  - New pure router `WheelDecision::{route, route_horizontal}` with `WheelConfig` and `WheelModifiers`. It replaces the commented-out alacritty-era router.
  - The unused `WheelDir` is removed.
  - `OrzmaTty::send_mouse` now writes nothing while the VT has no tracking level. A report routed on the GUI's slightly stale modes after nvim exits no longer reaches the shell as X10 bytes.
- **GUI.**
  - `src/input/mouse/wheel.rs` implements the routing above.
  - The local `WheelConfig` is replaced by the `orzma_tty` type.
  - `protocol_mods` moves to `src/input/mouse.rs` so the button and wheel paths share it.

Affected crates: `orzma_vt`, `orzma_tty`, and the root binary (`src/input`).

### Tests

- `orzma_vt`:
  - Alternate scroll defaults to enabled.
  - DECSET/DECRST 1007 select Enabled/Disabled.
  - Alternate scroll is active only on the alternate screen.
  - Mouse reporting is inactive only with tracking off.
  - RIS re-enables alternate scroll.
- `orzma_tty`:
  - 14 router tests cover priority, Shift fall-through, the burst cap before multiplying by lines, the 240-key limit, the fine modifier, zero notches, saturation, and horizontal routing.
  - The `send_mouse` guard is tested.
- GUI dispatch:
  - Reports per notch carry the cell under the cursor.
  - Alt is carried as the meta bit, and a held Super sets no bit.
  - The alternate screen receives cursor keys.
  - Shift over a tracking alternate screen sends cursor keys.
  - Horizontal orientation is checked.
  - The Shift fold runs on every OS in tests: line-unit travel folds, pixel-unit travel keeps its axes, and a folded frame clears the horizontal residual. The macOS end-to-end fold test is kept.
  - The fine-modifier helper is tested.
  - The existing dispatch tests pass unchanged.

### Manual verification (macOS)

- [ ] `nvim <long file>` scrolls with the wheel.
- [ ] `less <long file>` scrolls with the wheel (alternate scroll, no tracking).
- [ ] Back at the shell prompt, the wheel still scrolls the scrollback.
- [ ] A sideways swipe inside `nvim` does not leak vertical scroll.
- [ ] With both a discrete wheel and a trackpad:
  - Shift+wheel up over `nvim` moves toward the top of the file.
  - With `printf '\e[?1002h\e[?1006h'; cat -v`, scrolling toward the left side reports `^[[<66;…M` and the opposite way reports `^[[<67;…M`.


not-elm:
現状GUI側にターミナルのモードをミラーリングし、ホイール分岐をGUI側で行っているがこれは`orzma_tty`側で行った方がいいのではないか？
一度調査する必要がある。

PR #317, opened 2026-09-16
feat(orzma_tty, orzmux, input): route the mouse wheel inside the terminal by its live VT modes (PR-B of 2)

not-elm:
## Problem

The mouse wheel does nothing in nvim, `less`, or `man`. The wheel dispatcher (`src/input/mouse/wheel.rs`) always scrolls the pane's own scrollback, and the alternate screen has none, so the wheel is a dead key in every full-screen program. Wheel reports never reach applications that track the mouse.

#313 tried to fix this by mirroring each pane's VT modes into the GUI, but a mirrored value is always 0 to a few frames stale, and the arrow-key route has no way to re-check the live modes at apply time (wheel-derived arrow keys and real key presses reach the backend as the same `OrzmuxCommand::KeyInput`). This is PR-B of 2, stacked on #316, which removed that mirror. It replaces #314.

## Solution

The decision moves to the owner of the state. The GUI only normalizes the physical gesture and hands one plain-data `WheelInput` per frame to the terminal under the cursor; `OrzmaTty::send_wheel` reads its own VT modes once and routes both axes.

- **`orzma_vt`.** `AlternateScroll` (DECSET / DECRST 1007) becomes an enum, **enabled by default**; `VtModes::mouse_reporting_active()` and `alternate_scroll_active()`; the private-mode dispatch names 1000 / 1002 / 1003 / 1005 / 1006 explicitly instead of a catch-all (1005 stays a documented no-op).
- **`orzma_tty`.**
  - The pure router `WheelDecision::route` / `route_horizontal`: vertical notches become wheel reports while a tracking level is in force and Shift is not held, cursor keys while alternate scroll is in effect (honouring DECCKM), and a viewport scroll otherwise; horizontal notches become reports only. Reports are capped by `max_protocol_events_per_frame`, cursor keys additionally by `MAX_CURSOR_KEYS` (240).
  - `WheelInput { up, right, mods, cell, report_mods }` and `OrzmaTty::send_wheel(input, &WheelConfig)`: modes are read once, both axes are encoded into one buffer, and everything goes out in **one PTY write** (previously up to 32 per frame). A report is dropped when no cell was resolved; the cursor-key route snaps a scrolled-back viewport to the live tail, as `send_key` does.
  - `send_mouse` writes nothing while no tracking level is in force, so a report routed a moment before the app turned tracking off never leaks to the shell.
- **`orzmux`.** `OrzmuxCommand::Wheel { pane, input }`, a one-line delegation in `handle_command`, and `OrzmuxConfig.wheel: WheelConfig` handed to `Backend::new` at spawn (built from the `[mouse]` block in `main.rs`). Protocol purity holds: `WheelInput` is plain `Copy` data.
- **`bevy_orzmux`.** `RequestTtyWheel { terminal, input }` and its observer, an exact analogue of `RequestTtyMouseInput`.
- **GUI (`src/input`).** `dispatch_mouse_wheel` keeps target resolution, sub-notch accumulation, the dominant-axis lock, the macOS Shift fold, the fine modifier, and the cell projection, then fires one `RequestTtyWheel`; it no longer imports any VT mode type. The wheel-only `TerminalViewportScroll` action is deleted. `OrzmaMouseConfig` drops its `wheel` policy (the backend owns it now) and gains `OrzmaMouseConfig::from_config`. `protocol_mods` moves from `button.rs` to `mouse.rs`, shared by both call sites.
- **Config.** No `[mouse]` key changes. `max_protocol_events_per_frame` now caps notches per routing call (the backend has no frames); `docs/configs.md` says so.

Affected crates: `orzma_vt`, `orzma_tty`, `orzmux`, `bevy_orzmux`; binary: `src/input`, `src/configs.rs`, `src/main.rs`, `src/action/terminal`; docs: `CLAUDE.md`, `docs/configs.md`.

### Behavior changes

- **Wheel over a mouse-tracking app** (nvim with `mouse=nvi`, fzf) sends SGR / X10 wheel reports. **Shift + wheel** skips reporting and falls through to the next route, so it scrolls the terminal over fzf and sends arrow keys over nvim.
- **Alternate scroll is on by default.** Wheel over `less`, `man`, or any alternate-screen program that does not track the mouse now sends arrow keys to the program instead of scrolling the pane's (empty) scrollback. This matches alacritty, kitty, wezterm, and ghostty. Programs can turn it off with `DECRST 1007`; there is no config key.
- **Held Cmd / Super no longer sets the Meta bit** in mouse reports. xterm's bit 8 is Alt / Meta; Super never mapped to it, and the old code made a held Cmd indistinguishable from Alt.

### Tests

- `orzma_tty`: 14 router tests (each route, Shift fall-through, the fine modifier, both caps, DECCKM, zero notches) and 10 `send_wheel` tests that assert the PTY bytes: one report per notch in one write, the Alt bit, no cell → nothing written, `CSI A` / `SS3 A` on the alternate screen, Shift over a tracking alternate screen → cursor keys, horizontal → `WheelLeft` / `WheelRight`, primary screen → viewport scroll and a repaint, both axes in one write vertical first, the viewport snap only on the cursor-key route, and a PTY write failure surfacing as `PtyWrite`.
- `orzmux`: a `Wheel` command reaches its pane's terminal and is routed by that pane's own modes (DECSET 1002 / 1006 driven through real VT output); an unknown pane writes nothing and emits nothing; the backend's own `WheelConfig` (not the default) is what reaches `send_wheel`.
- `bevy_orzmux`: a `RequestTtyWheel` becomes a `Wheel` command carrying the same input; a non-pane entity sends nothing; no request observer runs without a connection.
- GUI: 13 tests on `dispatch_mouse_wheel` (notch counts and signs, the axis lock, the diagonal case with the lock off, the cell under the cursor, Alt reaching both the report bits and the fine modifier, the macOS Shift fold on both platforms) plus the pure fold / modifier helpers; the `[mouse]` → `WheelConfig` and → `OrzmaMouseConfig` mappings.
- `orzma_vt`: DECSET / DECRST 1007, reset and soft-reset semantics for `AlternateScroll`.

### Follow-ups left open on purpose

- One discrete wheel click is two notches at the default `cells_per_notch = 0.5`, so a tracking app receives two reports and a pager six arrow keys per click. This predates the PR on the viewport route; whether the report route should count physical detents is a separate decision.
- `WheelInput` carries the Shift bit in both `mods` and `report_mods`; the second is inert (Shift never reaches a report). A flatter wire shape is possible but changes the protocol type.
- The report route has no constant ceiling; only two deliberately absurd config values (`cells_per_notch = 0` plus a huge `max_protocol_events_per_frame`) can make one click a multi-megabyte write.
- This branch is written against `Pty::write_all`. If #315 (the PTY write queue) merges first, `send_wheel` moves to `enqueue_write` and the one-write assertions are re-derived.

Closes #314.


crates/orzma_tty/src/lib.rs
not-elm, line 1: テストケースが肥大化してきたためモジュールを分割したい。

not-elm (commented):
FB

crates/orzma_tty/src/lib.rs
not-elm: Split in fb6343b4: the inline module becomes `src/tests.rs` (shared fixtures) plus one file per operation under `src/tests/` — pump, coalescer, resize, scroll, input, wheel, focus, host — mirroring `orzma_vt`'s `interpreter/tests`. No test bodies changed; the crate still runs 149 tests and the workspace suite is green.

PR #325, opened 2026-09-19
feat(orzma_vt, orzma_tty, orzmux): hold frames back during a synchronized update (CSI ?2026) and answer DECRQM

not-elm:
## Problem

Two Tier 2 items from `docs/todo/vt-conformance-remaining.md`, and they only pay off together.

**Synchronized output (`CSI ?2026 h/l`) was unimplemented.** The number fell through
`set_private_modes`' catch-all, so an application that wrapped a redraw in BSU/ESU got no
atomicity: with the frame coalescer's 12 ms hard cap, the user saw the half-drawn screen anyway.
fzf issues the pair around every frame; nvim and tmux use it too.

**DECRQM (`CSI ? Ps $ p`, `CSI Ps $ p`) had no arm**, so the terminal never answered a mode
query. nvim and tmux both probe mode 2026 with DECRQM and enable synchronized output only on a
`1` or `2` reply, so 2026 on its own would have reached neither of them.

### Synchronized output: withhold the frame, keep parsing

The alternative was to hold back the bytes of an open update and parse them at ESU, which is what
alacritty and kitty do. This PR suppresses frame emission instead, because:

- ESU is detected by the real parser, so the 8-bit CSI form (`\x9b?2026l`) and multi-parameter
  spellings (`CSI ?25;2026l`) close an update. A byte scan for the exact 8-byte sequence — vte's
  approach, and its own comment says so — misses both.
- DSR and DECRQM replies are not delayed, which is what makes the startup probe above safe.
- `FrameTracker` already accumulates damage across a frame, so not emitting is atomic for free;
  no 2 MiB buffer and no second parse.

`Vt::interpret` now stops at the byte that closes an update and reports `consumed`, so the owner
takes the completed frame before the rest of the chunk is interpreted. Without that, an
application writing `…ESU BSU…` in one chunk would lose the screen it had just declared complete —
with continuous output it would degrade to a torn frame every 150 ms. The caller resubmits
`chunk[consumed..]`; a chunk that closes no update is still consumed whole.

`orzma_tty` owns every clock, as it already did for the coalescer:

- a 150 ms deadline anchored at the first BSU, which a repeated BSU does not extend;
- frames withheld while that deadline holds, including the pane's first (bootstrap) frame;
- a frame taken at each close, paced to at most one per 12 ms so a `BSU`/`ESU` toggle storm cannot
  turn 4 KiB of PTY output into thousands of frames;
- `flush_now` (resize, and closing a pane) pierces the hold, since a resize invalidates a
  half-drawn screen anyway and the last frame of a dying pane has to get out.

On timeout only painting resumes; the VT's mode register is left alone, so DECRQM keeps reporting
what the application set. That matches Contour, which specified the mode.

`OrzmaTty::pump` / `flush_now` now return one ordered `Vec<PumpItem>` rather than separate signal
and frame fields: a frame taken mid-chunk has to sit between the signals raised before and after
it, or the GUI can despawn a webview before it receives the frame that still lists its placement.
`orzmux` forwards that list in order through one traversal shared by all three of its call sites,
which also means an OSC 7 `CurrentDir` on the resize-flush path now updates the pane's working
directory as it always did on the pump path.

### DECRQM / DECRPM

`DeviceState::private_mode_report` and `ansi_mode_report` answer from the state the setters
already write — `0` not recognized, `1` set, `2` reset. Mode 6 reports the screen on show; 47,
1047 and 1049 all report whether the alternate screen is shown; 1000/1002/1003 report set only
for the tracking level in force; mode 12 reflects the blink DECSCUSR selected. The reply echoes
the mode number as sent rather than the `u16` the lookup saturates to, so a probe for a number
above 65535 is not answered about a different mode.

Modes 3, 40, 95, 1005, 1034 and 1048 answer `0`: this terminal accepts and ignores them, or they
name a stateless operation. That is alacritty's answer; xterm replies `2` for 3 and 40.
A sweep test walks every number 0–2100 in both directions and fails if a mode that changes
observable state answers `0`, so the two tables cannot drift apart silently — `MouseEncoding`
already carries a TODO that would have caused exactly that for 1005.

### Behavior worth knowing at review time

Accepted trade-offs of holding frames, all bounded by the 150 ms timeout:

- A user scroll or selection repaint inside an open update waits for the close. The grid holds the
  application's partial redraw, so any frame taken mid-update would leak it.
- A host-driven `mount` over the control socket (the ConPTY path) waits for the close on a pane
  with no PTY output. Its frame waits with it, so the two stay consistent.
- Signals that *remove* a webview (`WebviewUnmount`, `WebviewEvicted`, an in-place re-mount) reach
  the GUI while the frame reflecting them is held, so a webview can vanish over stale cells if an
  application stalls between the APC verb and its ESU. No first-party app emits 2026.
- An application killed inside an update leaves the VT's mode set, so the next program's first
  update is not held and its startup DECRQM answers `1`. It recovers at that program's first ESU.
- A placement stranded and re-mounted across a close is evicted and re-created rather than updated
  in place, which also drops app-declared webview focus. This already happens whenever a 4 KiB
  read boundary falls between the strand and the re-mount; a close is one more such boundary.

### One open question

The close-frame pacing measures against the last frame of *any* kind, including the torn frames
emitted after a timeout. While an application runs back-to-back updates each longer than 150 ms
with continuous output, almost every close is then paced away and every presented frame is torn —
the degradation the mid-chunk stop exists to prevent. Pacing against the last *close* frame
instead fixes it with the same frame-count bound (about 1–3 per pump). Left as written because the
design says "since the last emit"; say the word and I will change it here.

### Docs

`docs/memo/decset.md` gains a synchronized-output section and a DECRQM reply table, and states the
pacing and the trade-offs above. The two finished entries leave
`docs/todo/vt-conformance-remaining.md`.

### Testing

`cargo test -p orzma_vt -p orzma_tty -p orzmux` — 1196 / 182 / 70 passing. `cargo clippy
--all-targets` and `cargo check --workspace --all-targets` clean. Not exercised by hand in the GUI
against a real nvim or fzf yet.


crates/orzma_tty/src/lib.rs
```diff
@@ -619,29 +665,70 @@ impl<V: Vt> OrzmaTty<V> {
...
+                && !self.coalescer.emitted_within(Self::SYNC_EMIT_INTERVAL, now)
+            {
+                self.emit_frame(now);
+            }
+            if update.consumed == 0 || update.consumed > rest.len() {
+                warn!(
+                    consumed = update.consumed,
+                    len = rest.len(),
+                    "the VT broke the interpret contract; dropping the rest of the chunk"
+                );
```
not-elm: ここはResult型にしてErrを返したほうがいい？

crates/orzma_tty/src/lib.rs
```diff
@@ -619,29 +665,70 @@ impl<V: Vt> OrzmaTty<V> {
...
+                    .map(|signal| PumpItem::Signal(TtySignal::Vt(signal))),
+            );
+            self.pending_replies.extend(update.replies);
+            self.track_synchronized_update(now);
+            if update.synchronized_update_closed
+                && !self.coalescer.emitted_within(Self::SYNC_EMIT_INTERVAL, now)
+            {
+                self.emit_frame(now);
+            }
+            if update.consumed == 0 || update.consumed > rest.len() {
```
not-elm: consumed==0の時も警告を出す必要がある

not-elm (commented):
FB

crates/orzma_tty/src/lib.rs
not-elm: `Result` 化しました（b16dd67）。

`feed_chunk` は `OrzmaTtyResult` を返すようにして、回復できない境界である `drain_chunks` でログするようにしています。

```rust
ChunkPoll::Chunk(chunk) => {
    if let Err(error) = self.feed_chunk(&chunk) {
        warn!(%error, "dropping the rest of the chunk");
    }
}
```

そのチャンクの残りを捨てて次のチャンクへ進む挙動は従来どおりです。テスト用 seam の `feed_bytes` も `OrzmaTtyResult` を返すようにしたので、`FakeVt` が契約を破った場合にテストが黙って通ることはなくなりました。

crates/orzma_tty/src/lib.rs
not-elm: `consumed == 0` は元々 `||` で同じ `warn!` に入っていたので警告自体は出ていましたが、`consumed > rest.len()` と一緒くたになっていて区別がつきませんでした。b16dd67 で別々のエラーに分けました。

```rust
if update.consumed == 0 {
    return Err(OrzmaTtyError::VtConsumedNothing { len: rest.len() });
}
if update.consumed > rest.len() {
    return Err(OrzmaTtyError::VtConsumedBeyondChunk {
        consumed: update.consumed,
        len: rest.len(),
    });
}
```

分けた理由は、同じ「`Vt` が契約を破った」でも放置したときの結果が違うからです。

| 条件 | 防いでいるもの |
| - | - |
| `consumed == 0` | `rest` が縮まず `feed_chunk` が無限ループ → orzmux バックエンドスレッドがハング |
| `consumed > rest.len()` | 次行 `&rest[update.consumed..]` のスライス範囲外 panic |

なお `consumed == 0` は同梱の `OrzmaVt` では到達不能です。`OrzmaVt::interpret` が `consumed: 0` を返すのは空チャンクのときだけで、非空なら `Interpreter::parse` の `consumed += 1` が同期更新クローズの `break` より前にあるため必ず `1` 以上になります。`OrzmaTty<V: Vt>` がジェネリックなので、`OrzmaVt` 以外の `Vt` 実装が契約を破った場合にだけ起きる異常系という位置づけです。

`sync.rs` の契約テストも `[0, 9]` のループから variant ごとの2本に割って、返るエラーを assert するようにしました。

crates/orzma_tty/src/lib.rs
```diff
@@ -619,29 +674,78 @@ impl<V: Vt> OrzmaTty<V> {
...
+                return Err(OrzmaTtyError::VtConsumedNothing { len: rest.len() });
+            }
+            if update.consumed > rest.len() {
+                return Err(OrzmaTtyError::VtConsumedBeyondChunk {
+                    consumed: update.consumed,
+                    len: rest.len(),
+                });
+            }
+            rest = &rest[update.consumed..];
         }
```
not-elm, line 731: Vtトレイトのデフォルト実装に置き換えられないか？
つまりデフォルト実装の関数側でchunkごとのoutputを集約してtty側に最終的なInterpretOutputを送信する。

not-elm (commented):
FB

PR #326, opened 2026-09-19
fix(orzma_vt, bevy_orzma_tty_renderer): restore a configured caret style and move the blink phase off the GPU

not-elm:
## Problem

Running nvim once left the caret blinking forever afterwards. It was not "the caret always blinks" — a fresh pane was correct, and plain zsh never touches the cursor style. The blink appeared only *after* a program exited.

Two things had to line up:

1. nvim sends `ESC [ 0 SP q` (DECSCUSR `Ps=0`) as its teardown reset. Since neovim `ebc69b387` (PR #40249, merged 2026-06-15) it patches that sequence in regardless of terminfo, so our `Se=\E[2 q` is ignored. nvim v0.12.5 still does this; `a3eff66e3` on master has since reverted it for bare `xterm`.
2. `with_decscusr` mapped `0 | 1` to a blinking block, faithfully to xterm's ctlseqs and vt510.pdf p.251. Once set, nothing cleared it until `RIS`, `DECSTR`, `CSI 2 SP q` or `CSI ?12l` arrived.

The VT → frame → uniform → shader path was never at fault: the blink flag survived every hop and the shader already left a steady caret lit.

`Ps=0` had no way to mean "go back to what the user configured", because there was nothing configured to go back to. That is what this PR adds.

## Solution

A user-facing `[cursor]` section becomes the caret's resting state, and the blink decision moves off the GPU into one pure function.

**`orzma_configs` — the new section.** `style` (block / underline / bar), `blink_interval`, `blink_timeout`, `thickness`, `unfocused_hollow`. Modelled on `font.rs` rather than `inactive_pane.rs`, because the top-level `deny_unknown_fields` only guards section *names* — a typo inside the section needs the per-struct attribute to be caught.

`blink_interval = 0` holds the caret steady whatever a program asks for. That is the only switch for a non-blinking caret, and it deliberately sits in the renderer rather than the VT: a veto held in VT state is re-armed by the next reset, whereas one held at paint time is out of reach of `RIS`, `DECSTR`, `DECSCUSR` and DEC mode 12 alike. kitty (`cursor_blink_interval 0`) and WezTerm (`cursor_blink_rate = 0`) put it in the same place.

**`orzma_vt` — the VT restores a configured style.** `DeviceState` takes a plain-data `CursorPolicy { initial }` at construction, so a freshly spawned pane already shows the configured caret before the shell writes a byte. `DECSCUSR 0`, `DECSCUSR 7` and `RIS` restore it. `Ps=1` still selects the blinking block — only `0` and an omitted parameter changed meaning, matching what Alacritty's vte, kitty and ghostty each do.

`DECSTR` deliberately does *not*: vt510.pdf p.277 Table 5-9 and vt220.pdf p.85 Table 4-10 both list only the text cursor's enable, and DEC STD-070 puts user preference features outside a soft reset. `tput init` sends `DECSTR` on its own, so restoring there rewrote the caret on every invocation. foot and WezTerm leave it alone for the same reason. `RIS` keeps restoring, which vt510.pdf p.331 and DEC STD-070 D.2 Note 2 both describe as recalling the user's saved settings.

DEC mode 12 is always honored. An earlier revision gave the config a veto over it, which is what ghostty does; that was dropped once `blink_interval` could express the same intent without leaving the config and the running program fighting over one flag.

**`bevy_orzma_tty_renderer` — the phase leaves the GPU.** `CaretStroke::new` applies focus, hollow rendering and the blink phase in a fixed order where a caret an earlier rule hid can never come back, and `CaretPaint::new` rejoins the viewport cell. The shader loses its clock entirely: `time_seconds` is gone from the uniform and `fract(params.time_seconds)` with it, replaced at the same std140 offset by `cursor_thickness_phys` so the 336-byte layout is unchanged.

The packed word is a `PackedCursorStyle` bitflags value — `VISIBLE`, the two shape bits, `HOLLOW` at bit 4 — and reaches `u32` only where `TerminalParams` needs it. Folding the shape and the hollow flag into one `CaretStroke` enum makes a bar-shaped hollow caret unrepresentable; the shader picks its hit test from the shape bits before it branches on the hollow bit, so that pairing would have drawn a block outline over a bar's cell coverage. Over a wide glyph the outline suppresses the pair's inner edges, so a caret on a CJK character is one rectangle rather than two boxes with a seam down the middle.

**`src/input` — the phase counts from the last keystroke.** A press advances `LastKeyInstant`; a release does not. Typing resets the phase and forces the caret visible, and after `blink_timeout` of no typing the caret settles lit rather than dark. No scheduler: the phase is one comparison evaluated per frame, so the two-timer race Alacritty fixed in `7ea927ffc` cannot occur here.

### Review order

The branch is one logical stack and can be read in four passes:

- **`8af42579..8ebbd294`** — config plus VT. This alone stops the nvim latch. Note that `3d45cc87` adds the DEC mode 12 veto and `7e83f518` later removes it; reading the pair together saves re-deriving the argument.
- **`c11918b9..9e624fbf`** — paint policy, focus plumbing, uniform and shader.
- `e1bc820d`, `daa3d60b`, `ff3955d5` — review findings and a simplify pass.
- **`53358218..fac34141`** — the caret-paint refactors (`CaretStroke`, the `cursor` module split, `PackedCursorStyle`), the mode-12 and `blink`-key removal, a second review pass, and the steady-caret route.

### Verification

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check` and `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` are clean; CI is green on all nine checks.

The WGSL is not compiled by the test suite (`Shader::from_wgsl` only stores the source), so it was parsed and validated separately with naga 29.0.4, pinned from this workspace's `Cargo.lock`; the validator was sanity-checked against a deliberately broken copy first. `the_shader_cursor_bits_match_the_rust_constants` pins all five cursor constants against the Rust side, and was itself checked by breaking `CURSOR_SHAPE_BAR` in the shader and watching it fail.

### Known gaps, deliberately left

- **The shader changes have not been looked at on a real GPU.** The hollow outline's per-axis bound and the `CURSOR_VISIBLE` early return are both reasoned and reviewed, but WGSL is never compiled by `cargo test`, so nothing here is evidence about pixels.
- **Focus is inferred from the absence of `PaneInactiveStyle`**, a component that exists to answer "how much do I dim this pane". It correlates with focus only because `on_active_pane_changed` happens to move both together, and spawning new panes with `PaneInactiveStyle::default()` (a visual no-op) is what keeps the inference correct for the first frames. The honest fix is a positive marker owned by `bevy_orzmux`, which already owns `OrzmuxActivePaneChanged`. Left out because it spans two crates and reverses a decision in the design doc.
- **A webview holding keyboard focus inside the active pane still gets a blinking caret**, for the same reason — the renderer cannot see `KeyboardFocused` or `FocusedWebview`. Falls out of the fix above.
- **An unrecognised `style` word falls back to the default silently**, unlike every other config enum in the crate. `docs/configs.md` now records each silent clamp and revert, but the divergence from `keyboard.rs`'s `rejects_unknown_value` is real and worth settling.
- **A custom terminfo entry** (`TERM=orzma`, `Se=\E[0 q`) was split into its own design note. Its only remaining benefit is preserving a non-block configured style after nvim ≥0.13 exits, it does not work under tmux, and it costs every user the `TERM`-unknown problem on every SSH hop.


crates/bevy_orzma_tty_renderer/src/cursor.rs
```diff
@@ -0,0 +1,284 @@
...
+use std::time::Duration;
+
+/// Bit 4 of the packed `cursor_style` u32 — set when the caret is drawn
+/// as an outline rather than filled.
+///
+/// The bit is the renderer's own: the VT packs bits 0-3 and leaves this
+/// one clear.
+pub const CURSOR_HOLLOW_BIT: u32 = 16;
+
+const _: () = assert!(
```
not-elm: 削除して

crates/orzma_vt/src/screen/cursor.rs
```diff
@@ -23,17 +30,22 @@ pub struct Cursor {
 
 impl Cursor {
     /// Packs the style into one u32: bit 0 is [`CURSOR_VISIBLE_BIT`],
-    /// bits 1-2 carry the shape (Block `0`, Underline `1`, Bar `2`), and
-    /// bit 3 carries the blinking flag.
+    /// [`CURSOR_SHAPE_MASK`] carries the shape (Block `0`, Underline
+    /// `1`, Bar `2`), and [`CURSOR_BLINKING_BIT`] carries the blinking
+    /// flag.
     pub fn pack_cursor_style(&self) -> u32 {
```
not-elm: この関数とbit定数はrendererクレートに移動させて

crates/bevy_orzma_tty_renderer/src/cursor.rs
```diff
@@ -0,0 +1,284 @@
...
+    let interval_ms = interval.as_millis();
+    if interval_ms == 0 {
+        return true;
+    }
+    (elapsed.as_millis() / interval_ms).is_multiple_of(2)
+}
+
+/// The cursor drawing settings the renderer reads each frame.
+#[derive(Resource, Debug, Clone, Copy)]
+pub struct CursorRenderConfig {
```
not-elm: Configsクレート内の設定スキーマと名前が衝突するため改名したい

PR #330, opened 2026-09-20
fix(input): make webview mouse input work on Windows, and split the mouse gate it exposed

not-elm:
## Problem

Mouse input to an embedded webview did nothing on Windows; the same build worked on macOS.

On Windows CEF runs with `multi_threaded_message_loop`, so `bevy_cef` exposes its browser host only as `Res<BrowsersProxy>` and never calls `init_non_send::<Browsers>()` (`bevy_cef-0.12.0/src/webview.rs:143`). The router read `Option<NonSend<Browsers>>` unconditionally, so every pointer path resolved to `None` — silently, because of the `Option`.

Making that path live exposed two more defects:

- `MouseDisabled` meant three different things at once (IME composing / unfocused, vi mode, and "the cursor is over a webview rect so the terminal yields"). The first two suppress input; the third is the opposite. The router honoured none, so a click in vi mode still drove the page.
- Once the router honoured it, an IME composition suppressed pointer input to a focused inline webview — but orzma deliberately supports composing *into* a page (`src/input/ime.rs`). Mid-preedit the user could not click a candidate the page renders, and the terminal skipped the same surface, so the input reached nothing.

## Solution

**Confine the platform split.** A `CefMouse` `SystemParam` (`src/input/mouse/webview/cef_sink.rs`) holds `NonSend<Browsers>` off Windows and `Res<BrowsersProxy>` on it. No `Browsers`/`BrowsersProxy` reference survives elsewhere.

**Split the marker in three.**

| Marker | Set for | Read by |
|---|---|---|
| `TerminalMouseDisabled` | vi mode, IME composing, unfocused window | terminal dispatchers, hyperlink hover |
| `WebviewMouseDisabled` | vi mode, unfocused window, an IME composition **no inline webview owns** | the webview router |
| `MouseClaimedByWebview` | cursor over an interactive inline rect | terminal side stands down; router acts |

The router also resolves the topmost surface and *then* gates, rather than filtering — filtering inherited fall-through-to-the-pane-below semantics, which would have handed a suppressed pane's press to a lower pane's webview.

**Narrow the composition term rather than dropping it.** The webview gate is `!focused || in_vi_mode || (ime.is_composing() && focused_webview.0.is_none())`. `ImeState` is window-global with no owner, so dropping `is_composing()` outright would un-suppress every rect in every pane while composing into a *terminal*; a click there sets `FocusedWebview`, `read_ime_events` then drops the pending commit, and the preedit is silently lost. `WebviewMouseDisabled` implies `TerminalMouseDisabled` in every state.

**Release the focused webview on vi-mode entry** (`src/action/vi/mode.rs`). The gate made the off-rect click that clears inline focus unreachable, and `classify_key_batch` routes keys to the page before vi resolution — so vi mode was entered but inert, with no pointer route out.

Also: an in-flight press is released to CEF when suppression begins mid-press; the three router systems match their terminal-side counterparts on when they run; a `warn_once!` that could not fire is removed.

### Testing

`cargo test --workspace` 2477 passed / 0 failed; clippy `-D warnings` and `cargo fmt --check` clean. 18 tests added, including an end-to-end one that registers the real gate and move systems together and was confirmed red against the pre-fix router.

Seven router tests are `#[cfg(target_os = "windows")]` because `BrowsersProxy` is the only fakeable assertion channel, though the logic they pin is not Windows-specific.

**Not verified by hand on Windows** — no one has driven a real CEF window with a mouse on this branch.

### Follow-ups (not here)

- `mouse_leave: true` is never sent, so CEF hover state sticks after the cursor leaves a rect. This branch adds entry points to an existing defect.
- `src/action/vi/mode.rs` still inserts `TerminalMouseDisabled` directly, redundantly with the gate.
- `route_webview_pointer` and `forward_webview_mouse_moves` still use the `run_if(not(any_with_component::<GrabbedSeparator>))` form the wheel path moved away from, so one stale hover position is forwarded when a separator drag ends.
- `ImeState` has no owner entity; the `is_none()` conjunct approximates one.


src/input/mouse/webview/cef_sink.rs
```diff
@@ -0,0 +1,234 @@
...
+            return;
+        };
+        #[cfg(not(target_os = "windows"))]
+        sink.send_mouse_move(webview, buttons, position, mouse_leave);
+        #[cfg(target_os = "windows")]
+        {
+            let held: Vec<MouseButton> = buttons.into_iter().copied().collect();
+            sink.send_mouse_move(webview, &held, position, mouse_leave);
+        }
+    }
```
not-elm, line 77: `pub fn`にして

not-elm: Done in 9fb4388 — the four `CefMouse` methods are now `pub fn`. The struct keeps `pub(in crate::input::mouse)`, which caps the effective reachability, so this is a readability change rather than a widening. rustfmt collapsed `send_mouse_wheel` back onto one line now that the shorter visibility fits.

PR #332, opened 2026-09-21
perf(bevy_orzma_tty_renderer): refill terminal rows in place instead of reallocating them

not-elm:
## Problem

`TerminalCells::apply` rebuilt every row a frame carried: a fresh `Vec<GridSlot>` per row and a fresh `String` per cell. Output that scrolls marks the whole screen dirty, so a 200×50 pane paid 10,050 allocations on every such frame.

## Solution

`runs_to_cells` becomes `refill_row`, which refills the existing row in place and reuses each cell's text buffer. A reused buffer is trimmed back to 64 bytes so one oversized cluster cannot pin memory. `apply` now levels row widths after the refill, so a column-count change also keeps the buffers of the rows the frame carries.

On a 200×50 full redraw, re-applying a frame of the same shape now costs 0 allocations, down from 10,050 (measured with a counting allocator). The first frame into an empty grid still costs 10,051. The zero holds for rows whose shape does not change and whose text stays within the VT's own limits; the VT-side `Row::to_runs` allocations are unchanged.

Also removes the `docs/todo/renderer-cell-allocation.md` working note. The by-reference GPU upload from the same investigation is #333.

### Verification

`cargo test -p bevy_orzma_tty_renderer` (140 passed), `cargo clippy --all-targets`, `cargo fmt --check`, and `cargo build` are clean.


crates/bevy_orzma_tty_renderer/src/schema/grid.rs
```diff
@@ -338,54 +338,76 @@ impl TerminalCells {
...
-/// slots, resolving each run's hyperlink id against the retained table.
+/// Capacity, in bytes, above which a reused cell's text buffer is
+/// trimmed back before it is refilled.
+const RETAINED_TEXT_CAPACITY: usize = 64;
+
+const _: () = assert!(
+    (1 + MAX_COMBINING) * char::MAX_LEN_UTF8 <= RETAINED_TEXT_CAPACITY,
+    "the largest cell text the VT emits outgrows the retained text capacity",
+);
+
```
not-elm: 削除して

docs/todo/renderer-cell-allocation.md
not-elm, line 1: このファイルを削除して

not-elm (commented):
FB

